PR #2, opened 2020-06-25
Language analysis and some tests with proper CLA

rob-marcus:
Fixed author stuff from last time I tried to do this. Oops.


preprocess/analysis/language_analysis_tests.py
```diff
@@ -0,0 +1,111 @@
...
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Jun 23 16:36:57 2020
+
+@author: robert
+"""
+import unittest
+from language_analysis import analyze_entities, analyze_sentiment
+import json
+
```
kirakira0, line 11: Some discussion on the python style guide for ordering import statements: https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement
Looks like you don't **have** to alphabetize imports according to PEP, but it's kind of the standard

preprocess/analysis/language_analysis_tests.py
```diff
@@ -0,0 +1,111 @@
...
+    expected_entity_list = [("John Bel Edwards", "PERSON", 0.41),
+                            ("LA", "LOCATION", 0.29),
+                            ("No Sanctuary Cities", "LOCATION", 0.22),
+                            ("It's Up to You", "WORK_OF_ART", 0.09)]
+    formatted_expected_entity_list = entity_list(expected_entity_list)
+    
+    actual_list = format_entity_list(analyze_entities(STRING_TWO))
+
+    self.assertEqual(formatted_expected_entity_list, actual_list)
+
```
kirakira0, line 103: You might want to add some tests where the entities aren't formatted correctly and make sure those cases are handled appropriately

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
+"""
+@Author: Scarlet Nguyen 
+"""
+
```
kirakira0, line 4: A doc comment might be nice to explain where the mapping fits into the SA process

preprocess/analysis/language_analysis_tests.py
```diff
@@ -0,0 +1,111 @@
...
+    
+  Returns:
+    a list of dictionaries, where each dictionary maps name, type, salience
+    to their correspoding string/string/float value.
+  """
+  
+  return list(map(lambda x: json.loads(x), entity_list))
+
+class test_language_analysis(unittest.TestCase):
+  """Unit tests for language analysis
```
kirakira0, line 64: I don't know if you **need** the period here, but adding it would keep things consistent

preprocess/analysis/language_analysis_tests.py
rob-marcus: <3

preprocess/analysis/language_analysis_tests.py
rob-marcus: That's a good catch, who told you about that?!? But you should really know, per the google style guide, it's *necessary*.

rob-marcus: "A docstring should be organized as a summary line (one physical line) terminated by a period"

http://google.github.io/styleguide/pyguide.html#381-docstrings

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+  Args: 
+    number: float or an int.
+  
+  Returns:
+    a float to tens precisions.
+  """
+  
+  return float(round(number, 1))
+
+def round_to_2(number):
```
scarletnguyen13, line 36: round_to_1 and round_to_2 are duplicated, you can combine both of them to one single function with 2 args instead => something like `def round(number, toPlace)`

scarletnguyen13: Also, next time name the function a bit more clear than this, something like: round_to_hundredth and round_to_tenth

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+  
+  Returns: 
+    A json string mapping score and magnitude keys to their corresponding
+    value.
+    
+  Raises:
+    TypeError: input is not a string.
+  """
+  
+  if not isinstance(text, str):
```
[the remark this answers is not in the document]
scarletnguyen13: @rob-marcus to check for empty string before isinstance, you can have a condition of `if not (str and isinstance(text, str)):`

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+  Returns: 
+    A json string mapping score and magnitude keys to their corresponding
+    value.
+    
+  Raises:
+    TypeError: input is not a string.
+  """
+  
+  if not isinstance(text, str): 
+    raise TypeError("Input to analyze_sentiment isn't a string.")
```
scarletnguyen13, line 68: Include the error string here for clarification later

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+    
+  document = types.Document(content=text, type=enums.Document.Type.PLAIN_TEXT)
+
+  document_sentiment = \
+      CLIENT.analyze_sentiment(document,encoding_type=ENCODING_TYPE) \
+            .document_sentiment
+            
+  tens_place_score = round_to_1(document_sentiment.score)
+  tens_place_magnitude = round_to_1(document_sentiment.magnitude)
```
scarletnguyen13, line 77: nit: rounded_score and rounded_magnitude are sufficient enough

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+                                   "magnitude": tens_place_magnitude})
+
+  return score_and_magnitude
+
+def analyze_entities(text):
+  """Applies entity analysis to an input string.
+  
+  Calls cloud language API once for entity analysis if input is a string. 
+  Rounds salience value to the nearest hundreth for convenience. 
+
```
scarletnguyen13, line 89: Can you clarify here why we should round the score and magnitude to tenths while salience to hundredths?

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+    A list of json strings, where each json string maps 
+    name/type/salience keys to their corresponding string/string/float
+    values.
+    
+  Raises:
+    TypeError: input is not a string.
+  """
+  
+  if not isinstance(text, str): 
+    raise TypeError("Input to analyze_entities isn't a string.")
```
scarletnguyen13, line 103: Similar to above's comments!

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
...
+csv_file = csv.reader(file)
+
+count = 1
+
+existed = dict()
+
+for row in csv_file:
+  print(count)
+  if count > 36730:
```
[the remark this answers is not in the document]
scarletnguyen13: @rob-marcus can you remove this line? We don't need it anymore! :D

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def round_to_1(number):
+  """Rounds number to the nearest tenth. 
+
+  Cloud language api will automatically return floats to like 15 decimal 
+  places. It's convenient to just round down. 
+
+  Args: 
+    number: float or an int.
```
andrewlathamgoogle, line 28: Why would I provide an int to round to nearest tenth?

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+  document_entities = \
+      CLIENT.analyze_entities(document,encoding_type=ENCODING_TYPE) \
+            .entities
+  
+  formatted_entities = []
+  
+  for entity in document_entities: 
+    entity_name = entity.name
+    entity_type = enums.Entity.Type(entity.type).name
+    entity_salience = round_to_2(entity.salience)
```
andrewlathamgoogle, line 116: Is there a reason why salience is rounded to hundredth but score+magnitude rounded to tenth?

preprocess/analysis/language_analysis_tests.py
andrewlathamgoogle: Just use a python linter and you shouldn't have to worry about it

preprocess/analysis/language_analysis_tests.py
```diff
@@ -0,0 +1,111 @@
...
+  
+  Tests sentiment_analysis and entity_analysis with control values, and makes
+  sure that malformed inputs aren't accepted. 
+  """
+  
+  def test_string_one_sentiment(self):
```
andrewlathamgoogle, line 70: These names for your tests are not clear.

Please name tests:

test_method_scenario_expectedOutcome

All test descriptions should be related to the actual code, not your test file.  "Test String1" doesn't mean anything to the actual code.

preprocess/analysis/language_analysis_tests.py
andrewlathamgoogle: +1.  I'll pay you $1,000,000 to break this code.  How do you do it?

Think about that, then write those test cases.

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
+"""
+@Author: Scarlet Nguyen
```
andrewlathamgoogle, line 2: Why is the author Scarlet when you wrote this

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
+"""
```
andrewlathamgoogle, line 1: Not a very descriptive file name

preprocess/analysis/readme.txt
```diff
@@ -0,0 +1,8 @@
+To run sentiment analysis on your local machine you will to follow these instructions:
+https://cloud.google.com/natural-language/docs/setup
+
+in particular, you will need an api key. Don't push your api key to github please. 
+
+The test file with STRING_ONE and STRING_TWO runs for me in about 1 second, so expect about 1/4 of a second per api call. Query rate limit is 600/minute so that's pretty good.  
+
+Credit to Scarlet for putting together the scraping and connecting the analysis to the CSV, thank you scarlet!
```
andrewlathamgoogle, line 8: Scarlet's name deserves capitalization

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
...
+          analysis_info.append(new_dict.get('sentiment'))
+          analysis_info.append(new_dict.get('entities'))
+          existed[text] = new_dict
+
+      row.extend(analysis_info)
+  
+    with open('output.csv','a') as csv_text:
+      csvwriter = csv.writer(csv_text)
+      csvwriter.writerow(row)
+      print(analysis_info)
```
andrewlathamgoogle, line 47: Why are you printing this?  If it's just for debugging, remove.  This is going to dump millions of rows to your console and become a bottleneck when you actually run.

Same for print(count) above.

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
...
+      print('\n')
+  count+=1
```
andrewlathamgoogle, line 49: nit: spacing

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
...
+existed = dict()
+
+for row in csv_file:
+  print(count)
+  if count > 36730:
+    analysis_info = []
+    if count == 1:
+      row.extend(['Headline_Sentiment', 'Headline_Entities', 'Content_Sentiment', 'Content_Entities'])
+    else:
+      headlineAndContent = [row[len(row) - 3], row[len(row) - 1]]
```
andrewlathamgoogle, line 31: you need to add a comment explaining this

preprocess/analysis/mapping.py
```diff
@@ -0,0 +1,49 @@
...
+
+from language_analysis import analyze_entities, analyze_sentiment
+
+def get_analysis(text):
+  return {
+    'sentiment': analyze_sentiment(text),
+    'entities': analyze_entities(text)
+  }
+
+file = open('input.csv')
```
andrewlathamgoogle, line 16: Don't just dump your Python code as raw lines to be execute in the file, use if __name__ == '__main__' unless there's some reason not to

preprocess/analysis/mapping.py
rob-marcus: I've removed Scarlet's code from this PR, she is going to handle pushing this code. Going to mark these comments as resolved as Scarlet has seen them.

preprocess/analysis/readme.txt
rob-marcus: Done.

preprocess/analysis/mapping.py
rob-marcus: I've removed Scarlet's code from this PR, she is going to handle pushing this code. Going to mark these comments as resolved as Scarlet has seen them.

preprocess/analysis/language_analysis_tests.py
rob-marcus: Done.

preprocess/analysis/language_analysis.py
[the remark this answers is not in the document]
rob-marcus: Empty strings are valid inputs technically and will just return '{"score": 0.0, "magnitude": 0.0}' or []. 

The same goes for whitespace strings, i.e., `"\n\n\n\n\n"`, `" "*1000`, etc.  

Since these are acceptable inputs and outputs, I'm going to say they definitely aren't a `TypeError`. 

I think it's debatable if they are another type of error. If an ads blank/empty, I don't think the place to throw an error or not process that ad is in sentiment analysis. Probably somewhere else in the processing/parsing stream.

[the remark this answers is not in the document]
rob-marcus: The api will also handle non-`utf-8` encodings.

preprocess/analysis/language_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+  Returns: 
+    A list of json strings, where each json string maps 
+    name/type/salience keys to their corresponding string/string/float
+    values.
+    
+  Raises:
+    TypeError: input is not a string.
+  """
+  
+  if not isinstance(text, str):
```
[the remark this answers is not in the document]
rob-marcus: See discussion here
https://github.com/googleinterns/step9-2020/pull/2#discussion_r448584805

preprocess/analysis/language_analysis_tests.py
rob-marcus: See discussion here https://github.com/googleinterns/step9-2020/pull/2#discussion_r448584805

Tl;dr:
- the next level of useful testing is probably out of scope and would require some kind of language detection. 
- as long as the input is a valid python string, it's going to be a valid input for the API call.
- encoding  type is a non-issue.

preprocess/analysis/language_analysis.py
rob-marcus: Done, using the built in method and will just have a comment about sig figs.

preprocess/analysis/language_analysis.py
rob-marcus: Done.

preprocess/analysis/language_analysis.py
rob-marcus: Done. 

Tl;dr: I'm just following the display standard set by the api demo, so whatever rationale they had in deciding that seems reasonable to me.

preprocess/analysis/language_analysis.py
rob-marcus: Done.

preprocess/analysis/language_analysis.py
rob-marcus: Resolved by rewrite.

preprocess/analysis/language_analysis.py
rob-marcus: Done.

Tl;dr: I'm just following the display standard set by the api demo, so whatever rationale they had in deciding that seems reasonable to me.

rob-marcus: Api demo: 
https://cloud.google.com/natural-language#natural-language-api-demo

rob-marcus: Also changed the sentiment `score` rounding to the hundreds place because it seems like it can be that sensitive, though I've never seen it.

PR #3, opened 2020-06-29
Setup React config

scarletnguyen13:


scarletnguyen13:
> @googlebot I signed it!

PR #4, opened 2020-06-30
Complete frontend implementation in React and InstantSearch

scarletnguyen13:
Design:

Screen Shot 2020-07-01 at 1 29 16 PM


client/src/App.js
```diff
@@ -1,69 +1,8 @@
...
-        <Highlight attribute="name" hit={props.hit} />
-      </h1>
-    </article>
-  );
-}
-
-Hit.propTypes = {
-  hit: PropTypes.object.isRequired,
-};
+const App = () => <Search />;
```
rob-marcus, line 6: Nit: space

client/src/components/Hit/Hit.js
```diff
@@ -0,0 +1,23 @@
+import { Highlight } from 'react-instantsearch-dom';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+function Hit(props) {
```
rob-marcus, line 5: I think hits and props count as Algolia-domain specific, so JSdoc and some comments would be nice.

client/src/components/RangeSlider/RangeSlider.js
```diff
@@ -0,0 +1,22 @@
+import 'react-input-range/lib/css/index.css';
+import './RangeSlider.css';
+
+import React, { useState } from 'react';
+
+import InputRange from 'react-input-range';
+
+const RangeSlider = () => {
+  const [value, setValue] = useState({ min: 20000, max: 80000 });
```
rob-marcus, line 9: Comments for why _these_ magic numbers would be nice, and what they are supposed to reflect (guessing impressions?)

client/src/components/Search/Search.js
```diff
@@ -0,0 +1,81 @@
...
+                <div className="filter-item">
+                  <p className="filter-text">U.S States</p>
+                  <RefinementList
+                    attribute="data.geoTarget"
+                    limit={5}
+                    searchable
+                    showMore
+                  />
+                </div>
+                <div className="filter-item">
```
rob-marcus, line 50: Not sure if it's react-esque but the filter-item divs seems like you could write a function to generate/populate these more efficiently without hardcoding them.

client/src/constants/algolia_config.js
```diff
@@ -0,0 +1,8 @@
+import algoliasearch from 'algoliasearch/lite';
```
rob-marcus, line 1: why are we using algoliasearch/lite instead of regular Algolia?

client/src/constants/filter_constant.js
```diff
@@ -0,0 +1,19 @@
...
+  { label: '10K - 100K', start: 10001, end: 100000 },
+  { label: '100K - 1M', start: 100001, end: 1000000 },
+  { label: '1M - 10M', start: 1000001, end: 10000000 },
+  { label: '> 10M', start: 10000001 },
+];
+
+const COST = [
+  { label: 'All' },
+  { label: '≤ $100', start: 0, end: 100 },
+  { label: '$100 - $1,000', start: 101, end: 1000 },
```
rob-marcus, line 13: Aren't these bounds inclusive? Will it break if the start/ends overlap?

client/src/App.js
scarletnguyen13: it's not a React's convention!

client/src/constants/filter_constant.js
scarletnguyen13: Since my start and end boundaries are all separated clearly, we won't run into that problem

client/src/constants/algolia_config.js
scarletnguyen13: lite is much faster and we don't need the full functionality for MVP

client/src/components/Search/Search.js
scarletnguyen13: Great suggestion! Thank you I totally forgot about this haha

client/src/constants/filter_constant.js
rob-marcus: Misspoke. If the true range is 101 - 1000, it should be labeled that way.

client/src/constants/algolia_config.js
rob-marcus: Smart, good idea.

client/src/components/ResultItem/ResultItem.js
```diff
@@ -0,0 +1,34 @@
...
+
+  // Open url in a new tab
+  const handleLinkClicked = () => {
+    const url = `https://www.${hit.data.link}`;
+    const win = window.open(url, '_blank');
+    win.focus();
+  };
+  return (
+    <article>
+      <h1 onClick={handleLinkClicked} className="highlight-header">
```
andrewlathamgoogle, line 20: Easier (and more accessible) to just use a styled \<a\> element here.  Then you can do

target="_blank"
rel="noopener noreferrer"

which will avoid security issues that Safari loves to complain about

client/src/App.css
```diff
@@ -3,58 +3,23 @@ em {
...
-  margin-bottom: 2rem;
+html {
+  min-height: 100vh;
+  min-width: 100vw;
 }
 
-.pagination {
-  margin: 2rem auto;
-  text-align: center;
+@media screen and (max-width: 1050px) {
```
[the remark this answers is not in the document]
scarletnguyen13: It has always worked for me whenever I want to adjust different styles for different screen sizes, so I guess it works now?

client/src/styles/colors.css
```diff
@@ -0,0 +1,3 @@
+:root {
```
andrewlathamgoogle, line 1: Usually it's macros but this works too

client/src/components/ResultItem/ResultItem.css
```diff
@@ -1,7 +1,9 @@
+@import url("../../styles/colors.css");
+
 .highlight-header {
-  color: #0169a1 !important;
-  cursor: pointer;
-  text-decoration: underline;
+  color: var(--primary-color) !important;
```
andrewlathamgoogle, line 4: I don't think you need !important here

PR #7, opened 2020-07-01
DB Population

kirakira0:


server/functions/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
scarletnguyen13, line 1: 1. Hava another .gitignore file for the server/node_modules 
2. Remove server/node_modules and add-commit-push
3. npm install in your local machine to re-generate node_modules in server

server/node_modules/.bin/acorn
```diff
@@ -0,0 +1 @@
+../acorn/bin/acorn
```
scarletnguyen13, line 1: All node_modules' files below should be removed/git ignored

.gitignore
```diff
@@ -1,2 +1,3 @@
 **/.DS_Store
-**/.vscode
\ No newline at end of file
+**/.vscode
+test.txt
```
scarletnguyen13, line 3: why is the test.txt still in this PR although it's in the gitignore file?

server/pom.xml
```diff
@@ -7,7 +7,14 @@
   <groupId>com.google.sps</groupId>
   <artifactId>server</artifactId>
   <version>1</version>
-  <packaging>war</packaging>
+  <packaging>jar</packaging>
+
+  <!-- <repositories>
```
scarletnguyen13, line 12: avoid leaving comments in complete PR - do you still need this dependency?

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+import java.lang.StringBuilder;
+import java.time.LocalDate;
+import java.util.Arrays;
+import java.util.List; 
+
+public class Ad {
+
+    public String id; 
+    public String advertiser;  
+    public String startDate;
```
scarletnguyen13, line 17: add a comment here verifying that the date is specifically in ISO String

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+    public String endDate;
```
scarletnguyen13, line 18: same for startDate

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+    public long impressionsMin; 
+    public long impressionsMax; 
+    public boolean ageTargetingEnabled;
```
scarletnguyen13, line 21: nit: a better name could be `isTargetingAge`

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+    public long spendMax; 
+    public String headline; 
+    public String link;
+    public String content;  
+    public String headlineSentiment; 
+    public String headlineTerms;
+    public String contentSentiment; 
+    public String contentTerms; 
+
+  public Ad() {
```
scarletnguyen13, line 34: I think it's okay to write it like this: `public Ad() {}`, then put the comment right above the constructor

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+      .append("\nHeadline Sentiment: " + this.headlineSentiment) 
+      .append("\nHeadline Terms: " + this.headlineTerms) 
+      .append("\nContent Sentiment: " + this.contentSentiment) 
+      .append("\nContent Terms: " + this.contentTerms) 
+      .toString(); 
+    return s; 
+  }
+
+  public static class AdBuilder {
+
```
scarletnguyen13, line 85: If AdBuilder and Ad object both have the same long list of fields, is there any way that we could reduce this duplication? Something like the Ad object contains only 1 field called AdBuilder that contains all props instead having the exact same fields as the AdBuilder?

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+    public String headlineTerms;
+    public String contentSentiment; 
+    public String contentTerms; 
+
+  public Ad() {
+  // Must have a public no-argument constructor for Firesbase
+  }  
+
+  private Ad(AdBuilder builder) {
+    this.id = builder.id;
```
scarletnguyen13, line 39: is something like `this.builder = builder` feasible?

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+    public String headline; 
+    public String link;
+    public String content;  
+    public String headlineSentiment; 
+    public String headlineTerms;
+    public String contentSentiment; 
+    public String contentTerms; 
+
+    public AdBuilder id(String adId) {
+      this.id = adId;
```
scarletnguyen13, line 106: this method's name is not clear as it doesn't reveal the function's intention (set the field name and return the whole object).

=> Change the function's name to something like `setId(String adId)` and have a separate method to return the whole AdBuilder like `AdBuilder.getInstance()` (remember we had something similar for Datastore?)

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+public class AdRowProcessor {
+
+  private String[] row; 
+
+  public AdRowProcessor(String[] csvRow) {
+    this.row = csvRow; 
+  }
+
+  public Ad createAd() {
+    Ad ad = new Ad.AdBuilder()
```
scarletnguyen13, line 35: Why can't we directly create an Ad object but using an AdBuilder here?

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+      .headlineTerms(row[14].trim())
+      .contentSentiment(row[15].trim())
+      .contentTerms(row[16].trim())
+      .build();
+    System.out.println(ad.toString());
+    return ad;
+  }
+
+  public void addAdToDatabase(Ad ad, int rowIndex, String COLLECTION) throws Exception {
+    FileInputStream serviceAccount = new FileInputStream("../resources/accountKey.json");
```
scarletnguyen13, line 60: Extract the path `../resources/accountKey.json` to a constant and place it at the top of the file - so we can easily change it later - paths are extremely easy to misspell/misplace

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+    FirebaseOptions options = new FirebaseOptions.Builder()
+      .setCredentials(GoogleCredentials.fromStream(serviceAccount))
+      .setDatabaseUrl("https://step9-2020-capstone.firebaseio.com")
+      .build();
+    // initialize only if necessary 
+    if(FirebaseApp.getApps().isEmpty()) {
```
scarletnguyen13, line 66: Init the App in the constructor instead of here

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+  public long getImpressionsMax(String str) {
+    if (str.isEmpty()) {
+      return -1; 
+    }
+    String s = str.replace("k", "000").replace("M", "000000").replace(" ","").replace("\u2264", "");
+    String[] arr = s.split("-");
+    return Long.parseLong(arr[arr.length - 1]);   
+  }
+
+  public boolean isAgeTargetingEnabled(String str) {
```
scarletnguyen13, line 96: change function's name if you decided to rename the field's name as I suggested above

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+    if (str.trim().equals("Not targeted")) {
+      return false; 
+    }
+    return true; 
+  }
+
+  public List<String> getList(String str) {
+    List<String> l = Arrays.asList(str.split(","));
+    List<String> trimmedList = new ArrayList<String>();  
+    for(String s: l) {
```
scarletnguyen13, line 106: consider using Java Streams, it's super short and sweet, you'll fall in love with Streams as you get the hang of it ;)

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Description: File for reading in rows of CSV data
+ * Author: Kira Toal
+ * Date: June 24, 2020
+ */
+package com.google.sps.utils;
+
+import com.google.sps.utils.Ad;
+import com.google.sps.utils.AdRowProcessor;
+import java.io.*;
```
scarletnguyen13, line 10: Java guideline 3.3.1 - No wildcard imports. You must list out explicitly what subclasses you're using in this file

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,56 @@
...
+    try {    
+      BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(csvFile));      
+      String[] currentRow = {}; // array stores single row of CSV data (one advertisement)
+      // read header rows separately
+      String primaryHeader = bufferedReader.readLine();
+      String secondaryHeader = bufferedReader.readLine();
+      int rowIndex = 1; 
+      String row = ""; 
+      while((row = bufferedReader.readLine()) != null && (rowIndex < END_ROW_INDEX)) {
+        currentRow = row.split(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", -1);
```
scarletnguyen13, line 33: Add a short comment explaining the regex here

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,56 @@
...
+      String[] currentRow = {}; // array stores single row of CSV data (one advertisement)
+      // read header rows separately
+      String primaryHeader = bufferedReader.readLine();
+      String secondaryHeader = bufferedReader.readLine();
+      int rowIndex = 1; 
+      String row = ""; 
+      while((row = bufferedReader.readLine()) != null && (rowIndex < END_ROW_INDEX)) {
+        currentRow = row.split(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", -1); 
+        for (int currentColumn = 0; currentColumn < currentRow.length; currentColumn++) {
+          currentRow[currentColumn] = currentRow[currentColumn].replace("\"", ""); // remove extra quotes
```
scarletnguyen13, line 35: you can also handle this with single-quote, `replace('"', '')`

server/src/main/java/com/google/sps/utils/test.txt
```diff
@@ -0,0 +1 @@
+test file
```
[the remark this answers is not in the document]
kirakira0: No, I've removed it

.gitignore
kirakira0: I'm trying to figure that out too. Adding the filename to the .gitignore doesn't seem to change anything for me and I'm not sure why.

server/pom.xml
kirakira0: No, I've removed the unused dependencies

server/src/main/java/com/google/sps/utils/Ad.java
kirakira0: Got it!

server/firestore.rules
```diff
@@ -0,0 +1,8 @@
+rules_version = '2';
+service cloud.firestore {
+  match /databases/{database}/documents {
+    match /{document=**} {
+      allow read, write: if true;
```
rob-marcus, line 5: For future ref can you just make a comment about why we need this and what to change if we want to revert.

server/functions/index.js
```diff
@@ -0,0 +1,8 @@
+const functions = require('firebase-functions');
+
+// // Create and Deploy Your First Cloud Functions
+// // https://firebase.google.com/docs/functions/write-firebase-functions
+//
+// exports.helloWorld = functions.https.onRequest((request, response) => {
+//  response.send("Hello from Firebase!");
+// });
```
rob-marcus, line 8: Don't commit commented out code

server/src/main/java/com/google/sps/utils/Ad.java
kirakira0: That is a much better name, thank you

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+    public boolean ageTargetingEnabled; 
+    public List<String> genderTarget; 
+    public List<String> geoTarget;
+    public long spendMin; 
+    public long spendMax; 
+    public String headline; 
+    public String link;
+    public String content;  
+    public String headlineSentiment; 
+    public String headlineTerms;
```
rob-marcus, line 30: Isn't this a `List<String>` or is the list parsed from the csv as a string? If so I think it should be parsed in java as a string list for later usage. Same for `contentTerms`. 

Also please update the design doc schema when you have a minute to reflect the final state.

server/src/main/java/com/google/sps/utils/Ad.java
kirakira0: Cool!

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,201 @@
...
+    this.headlineSentiment = builder.headlineSentiment;
+    this.headlineTerms = builder.headlineTerms; 
+    this.contentSentiment = builder.contentSentiment; 
+    this.contentTerms = builder.contentTerms; 
+  }
+
+  @Override 
+  public String toString() {
+    String s = new StringBuilder()
+      .append("ID: " + this.id)
```
rob-marcus, line 62: Nit: I think this should be a 4 space indent, not a 2. 


https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent

server/src/main/java/com/google/sps/utils/Ad.java
rob-marcus: +1

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+      .spendMax(getLong(row[9]))
+      .headline(row[10].trim())
+      .link(row[11].substring(3)) // trim "Ad" from "Ad {URL}"
+      .content(row[12].trim())
+      .headlineSentiment(row[13].trim())
+      .headlineTerms(row[14].trim())
+      .contentSentiment(row[15].trim())
+      .contentTerms(row[16].trim())
+      .build();
+    System.out.println(ad.toString());
```
rob-marcus, line 55: why print this out every time we get the ad? If for debugging should be deleted.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+      .setCredentials(GoogleCredentials.fromStream(serviceAccount))
+      .setDatabaseUrl("https://step9-2020-capstone.firebaseio.com")
+      .build();
+    // initialize only if necessary 
+    if(FirebaseApp.getApps().isEmpty()) {
+      FirebaseApp.initializeApp(options);
+    }
+    Firestore db = FirestoreClient.getFirestore();    
+    ApiFuture<WriteResult> result = db.collection(COLLECTION).document(ad.id).set(ad);
+    System.out.println("Update time : " + result.get().getUpdateTime());
```
rob-marcus, line 71: Is this for debugging?

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public long getImpressionsMin(String str) {
+    if (str.isEmpty()) {
+      return -1; 
+    }
+    String s = str.replace("k", "000").replace("M", "000000").replace(" ","").replace("\u2264", "");
+    String[] arr = s.split("-");
+    // If csv field uses "<=", arr has length 1 and min number of impressions is 0.
+    if (arr.length > 1) {
+        return Long.parseLong(arr[0]);
```
rob-marcus, line 82: I think it would be smart to put this in a `try catch` statement in case the input is unexpectedly malformed etc. `parseLong` will raise a `numberFormatException` if it can't parse the input.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+    return 0;
+  }
+
+  public long getImpressionsMax(String str) {
+    if (str.isEmpty()) {
+      return -1; 
+    }
+    String s = str.replace("k", "000").replace("M", "000000").replace(" ","").replace("\u2264", "");
+    String[] arr = s.split("-");
+    return Long.parseLong(arr[arr.length - 1]);
```
rob-marcus, line 93: A lot of repetition with `getImpressionsMin`. Not sure if it's worth it here but you could probably treat these as wrappers for a more flexible version.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+    Firestore db = FirestoreClient.getFirestore();    
+    ApiFuture<WriteResult> result = db.collection(COLLECTION).document(ad.id).set(ad);
+    System.out.println("Update time : " + result.get().getUpdateTime());
+  }
+
+  public long getImpressionsMin(String str) {
+    if (str.isEmpty()) {
+      return -1; 
+    }
+    String s = str.replace("k", "000").replace("M", "000000").replace(" ","").replace("\u2264", "");
```
rob-marcus, line 78: kind of obvious because i've read the csv, but with magic numbers/strings it's best to leave a comment addressing them. a javadoc explaining the various options would be even better.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public boolean isAgeTargetingEnabled(String str) {
+    if (str.trim().equals("Not targeted")) {
+      return false; 
+    }
+    return true; 
+  }
+
+  public List<String> getList(String str) {
+    List<String> l = Arrays.asList(str.split(","));
```
rob-marcus, line 104: bad name

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+      trimmedList.add(s.trim());
+    }
+    return trimmedList; 
+  }
+
+  public long getLong(String str) {
+    if (str.isEmpty()) {
+      return -1; 
+    }
+    return (long)Double.parseDouble(str);
```
rob-marcus, line 116: Same as above with `NumberFormatException`.

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,56 @@
...
+  private static final int END_ROW_INDEX = 4; // last csv row (exclusive)
+
+  public static void readCSV(InputStream csvFile) throws Exception {
+    try {    
+      BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(csvFile));      
+      String[] currentRow = {}; // array stores single row of CSV data (one advertisement)
+      // read header rows separately
+      String primaryHeader = bufferedReader.readLine();
+      String secondaryHeader = bufferedReader.readLine();
+      int rowIndex = 1;
```
rob-marcus, line 30: Why is this 1-indexed?

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,56 @@
...
+      int rowIndex = 1; 
+      String row = ""; 
+      while((row = bufferedReader.readLine()) != null && (rowIndex < END_ROW_INDEX)) {
+        currentRow = row.split(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", -1); 
+        for (int currentColumn = 0; currentColumn < currentRow.length; currentColumn++) {
+          currentRow[currentColumn] = currentRow[currentColumn].replace("\"", ""); // remove extra quotes
+        }
+        // process the row
+        if (rowIndex >= START_ROW_INDEX) {
+          System.out.println("\nReading row #" + rowIndex + " ...\n");
```
rob-marcus, line 39: Consider deleting this print statement.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Got it

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Got it!

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Got it, initialized app in constructor instead

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Thanks for the recommendation! Adding streams made the method a lot less verbose

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: Got it

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: Got it

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Changed name

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: Got it, changed to start row index

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: Deleted

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+    FirebaseOptions options = new FirebaseOptions.Builder()
+        .setCredentials(GoogleCredentials.fromStream(serviceAccount))
+        .setDatabaseUrl("https://step9-2020-capstone.firebaseio.com")
+        .build();
+    if(FirebaseApp.getApps().isEmpty()) {
+      FirebaseApp.initializeApp(options);
+    }
+  }
+
+  public void addAdToDatabase(Ad ad, int rowIndex, String COLLECTION) throws Exception {
```
scarletnguyen13, line 47: May I ask why the `COLLECTION` is all-caps?

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+  private static final String PATH_TO_SERVICE_ACCOUNT = "./serviceAccountKey.json"; 
+  private String[] row; 
+
+  public AdRowProcessor(String[] csvRow) throws Exception {
+    this.row = csvRow; 
+    // initialize app
+    FileInputStream serviceAccount = new FileInputStream(PATH_TO_SERVICE_ACCOUNT);
+    FirebaseOptions options = new FirebaseOptions.Builder()
+        .setCredentials(GoogleCredentials.fromStream(serviceAccount))
+        .setDatabaseUrl("https://step9-2020-capstone.firebaseio.com")
```
scarletnguyen13, line 40: Put URL in a constant on top of class as well

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,59 @@
...
+        // process the row
+        if (rowIndex >= START_ROW_INDEX) {
+          AdRowProcessor processor = new AdRowProcessor(currentRow);
+          Ad ad = processor.createAd(); 
+          processor.addAdToDatabase(ad, rowIndex, COLLECTION); 
+        }
+        rowIndex++;  
+      }
+      bufferedReader.close();
+    } catch(IOException e) {
```
scarletnguyen13, line 50: if you already had `throws IOException,Exception` in `main`, this `catch` block is not necessary -> just throws the error in the method signature

server/functions/.gitignore
kirakira0: Ok, I have done so. I'm marking this conversation as resolved because the change is in another file and won't be shown as "outdated", and I want to indicate that I've addressed it.

.gitignore
kirakira0: Marking this as resolved to show that it's been addressed.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: I'm not sure why this isn't marked as outdated. I checked AdRowProcessor and the app is indeed being initialized in the constructor instead. I am marking this as resolved to show it's been addressed.

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: Again, I am unsure as to why this is not marked as outdated. There is a comment explaining regex in ReportReader. I am marking this as resolved to show that it's been addressed.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Got it

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        .contentTerms(row[16].trim())
+        .build();
+    return ad; 
+  }
+
+  public long getImpressionsMin(String str) {
+    if (str.isEmpty()) {
+      return -1; 
+    }
+    String s = str.replace("k", "000").replace("M", "000000").replace(" ","").replace("\u2264", "");
```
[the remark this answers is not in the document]
kirakira0: Got it

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        .map(String::trim)
+        .collect(Collectors.toList());
+    return trimmedList;  
+  }
+
+  public long getLong(String str) {
+    if (str.isEmpty()) {
+      return -1; 
+    }
+    return (long)Double.parseDouble(str);
```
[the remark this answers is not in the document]
kirakira0: Got it

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Got it, I added comments and a helper method for parsing the impressions

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: The result.get() is necessary for writing to firestore. I think I would like to keep the print statement in for now unless it slows down the process too much. From what I've seen it doesn't make much of an impact on the time it takes to update the db.

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,225 @@
...
+    private Ad instance; 
+    public Builder() {
+      instance = new Ad(); 
+    }
+    public Ad.Builder id(String id) {
+      instance.id = id; 
+      return this; 
+    }
+    public Ad.Builder advertiser(String advertiser) {
+      instance.advertiser = advertiser;
```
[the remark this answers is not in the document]
kirakira0: They're not really setters. I think it's customary to leave out the set keyword so you can add info to the builder like this:

Ad ad = Ad.newBuilder()
        .id(row[0])
        .advertiser(row[1])
        .startDate(row[2]) 
        .endDate(row[3])

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: Got it

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Done

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  public List<String> convertStringToList(String str) {
+    List<String> trimmedList = Arrays.stream(str.split(","))
+        .map(String::trim)
+        .collect(Collectors.toList());
+    return trimmedList;  
+  }
+
+  public long getLong(String str) {
```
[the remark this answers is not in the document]
kirakira0: Good point

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,225 @@
...
+  public void setHeadline(String headline) {
+    this.headline = headline; 
+  }
+  public void setLink(String link) {
+    this.link = link; 
+  }
+  public void setContent(String content) {
+    this.content = content; 
+  }
+  public void setHeadlineSentiment(String headlineSentiment) {
```
andrewlathamgoogle, line 71: For a future PR, consider making a set of Objects to encapsulate some of this information so it's not a whole bunch of isolated fields.  It seems like there is a logical grouping of some of these fields (like text, sentiment, terms -- an object representing this set of fields could be applied to both headline and content).

server/src/main/java/com/google/sps/utils/Ad.java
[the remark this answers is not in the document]
andrewlathamgoogle: I think what Scarlet means is why don't you do instance.setAdvertiser(advertiser) which I agree with.

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,225 @@
+/* 
+ * Author: Kira Toal
+ * Date: June 24, 2020
+ */
+package com.google.sps.utils;
+import java.util.List; 
+ 
+public class Ad {
+  // fields
```
andrewlathamgoogle, line 9: unnecessary comment (also below)

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,225 @@
...
+  public String id;
```
andrewlathamgoogle, line 10: Make these private so you control access to them through the getters and setters.

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,225 @@
+/* 
+ * Author: Kira Toal
+ * Date: June 24, 2020
+ */
+package com.google.sps.utils;
+import java.util.List; 
+ 
+public class Ad {
```
andrewlathamgoogle, line 8: Javadoc

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,225 @@
...
+  // fields
+  public String id; 
+  public String advertiser;  
+  public String startDate; // isostring format
+  public String endDate; //isostring
+  public long impressionsMin; 
+  public long impressionsMax; 
+  public boolean isTargetingAge; 
+  public List<String> genderTargets;
```
andrewlathamgoogle, line 17: It would be helpful to add more comments for the expected format for each of these, or what some of them are.   You could do this with javadocs on the getters+setters.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class AdRowProcessor {
+
+  private static final String PATH_TO_SERVICE_ACCOUNT = "./serviceAccountKey.json"; 
+  private static final String DATABASE_URL = "https://step9-2020-capstone.firebaseio.com"; 
+  private String[] row; 
+
+  public AdRowProcessor(String[] csvRow) throws Exception {
+    this.row = csvRow; 
+    // initialize app
```
andrewlathamgoogle, line 37: In general, when using comments to separate sections, add a new line before the comment.

Also capitalize the first letter of the comment and put a period at the end.

Also elsewhere.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+
+  public void addAdToDatabase(Ad ad, int rowIndex, String collection) throws Exception {
+    Firestore db = FirestoreClient.getFirestore();    
+    ApiFuture<WriteResult> result = db.collection(collection).document(ad.id).set(ad);
+    System.out.println("Update time : " + result.get().getUpdateTime());
+  }
+
+  public Ad createAd() {
+    Ad ad = Ad.newBuilder()
+        .id(row[0])
```
andrewlathamgoogle, line 56: make constants with what the index for each item in the row should be.  That way if this changes later it is easy to update.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
andrewlathamgoogle: I'll leave it up to you but I think you'll find that this becomes a bottleneck when you are trying to process a lot of rows at once and outputting this line every single time.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+    FirebaseOptions options = new FirebaseOptions.Builder()
+        .setCredentials(GoogleCredentials.fromStream(serviceAccount))
+        .setDatabaseUrl(DATABASE_URL)
+        .build();
+    if(FirebaseApp.getApps().isEmpty()) {
+      FirebaseApp.initializeApp(options);
+    }
+  }
+
+  public void addAdToDatabase(Ad ad, int rowIndex, String collection) throws Exception {
```
andrewlathamgoogle, line 48: writeAd is easier to say.  Also more accurate since this isn't just add, right, it's also update.

andrewlathamgoogle, line 48: This method is out of place in this file.  This file is all about processing the row and converting between row format and Ad format.  Yet you have this data access method here.  Move this to its own file.  Then you could get rid of the constructor and make this whole class a util class with static methods.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+    Firestore db = FirestoreClient.getFirestore();    
+    ApiFuture<WriteResult> result = db.collection(collection).document(ad.id).set(ad);
+    System.out.println("Update time : " + result.get().getUpdateTime());
+  }
+
+  public Ad createAd() {
```
andrewlathamgoogle, line 54: public static Ad convertRowToAd(csvRow)

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+        .headlineTerms(row[14].trim())
+        .contentSentiment(row[15].trim())
+        .contentTerms(row[16].trim())
+        .build();
+    return ad; 
+  }
+
+  /* Min and max impression fields contain letters k and M as well as ≤. The 
+  formatImpressionField helper method makes the fields easier to parse later.
+  */
```
andrewlathamgoogle, line 80: Format this javadoc comment properly

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+        .contentTerms(row[16].trim())
+        .build();
+    return ad; 
+  }
+
+  /* Min and max impression fields contain letters k and M as well as ≤. The 
+  formatImpressionField helper method makes the fields easier to parse later.
+  */ 
+  public String[] formatImpressionField(String str) {
+    String impressionsString = str.replace("k", "000").replace("M", "000000").replace(" ","").replace("≤", "");
```
andrewlathamgoogle, line 82: Future PR:  Include some validation in this and below methods to ensure that the format is always what you expect.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+    String[] impressionsArray = impressionsString.split("-");
+    return impressionsArray; 
+  }
+
+  public long getImpressionsMin(String impressionsField) throws IllegalArgumentException{
```
andrewlathamgoogle, line 87: nit: space between Exception and {

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+    String[] impressionsArray = formatImpressionField(impressionsField); 
+    // If csv field uses "<=", arr has length 1 and min number of impressions is 0.
+    if (impressionsArray.length > 1) {
+      try {
+        return Long.parseLong(impressionsArray[0]);
+      } catch (IllegalArgumentException e) {
```
andrewlathamgoogle, line 93: Does this throw an IllegalArgumentException?  I thought it was a NumberFormatException.

Anyway you should probably crash if this happens, not set it to a misleading value.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  public List<String> convertStringToList(String str) {
+    List<String> trimmedList = Arrays.stream(str.split(","))
+        .map(String::trim)
+        .collect(Collectors.toList());
+    return trimmedList;  
+  }
+
+  public long convertStringToLong(String str) {
```
andrewlathamgoogle, line 123: Is this expected in the data?

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,55 @@
...
+  public static void readCSV(InputStream csvFile) throws Exception {
+    BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(csvFile));      
+    String[] currentRow = {}; // array stores single row of CSV data (one advertisement)
+    // read header rows separately
+    String primaryHeader = bufferedReader.readLine();
+    String secondaryHeader = bufferedReader.readLine();
+    int rowIndex = START_ROW_INDEX; 
+    String row = ""; 
+    while((row = bufferedReader.readLine()) != null && (rowIndex < END_ROW_INDEX)) {
+      /* Regex splits the row into a String array by columns (columns are indicated in
```
andrewlathamgoogle, line 31: formatting

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+
+public class ReportReader {
+
+  private static final String CSV_FILE_PATH = "/output.csv";
+  private static final String COLLECTION = "testing"; 
+  private static final int START_ROW_INDEX = 1;  // first csv row that gets read in to DB (1 based)
```
andrewlathamgoogle, line 19: Is this because it has a header row?  Take a look and see if you can specify that in the parameters to InputStreamReader.

Also specify whether this is inclusive or exclusive.

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,55 @@
...
+        AdRowProcessor processor = new AdRowProcessor(currentRow);
+        Ad ad = processor.createAd(); 
+        processor.addAdToDatabase(ad, rowIndex, COLLECTION); 
+      }
+      rowIndex++;  
+    }
+    bufferedReader.close();
+  }
+
+  public static void main(String[] args) throws IOException,Exception {
```
andrewlathamgoogle, line 51: nit: space between , and Exception

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,55 @@
...
+    String secondaryHeader = bufferedReader.readLine();
+    int rowIndex = START_ROW_INDEX; 
+    String row = ""; 
+    while((row = bufferedReader.readLine()) != null && (rowIndex < END_ROW_INDEX)) {
+      /* Regex splits the row into a String array by columns (columns are indicated in 
+      the csv by commas.) It ignores commas inside quotation marks to avoid spliting 
+      individual text fields (the text field "North Carolina, Florida" is not split 
+      into two different Strings). 
+      */ 
+      currentRow = row.split(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", -1);
```
andrewlathamgoogle, line 36: Regexes are evil.  Please try to find a better way to do this.  It's ok if it takes a lot more code if it's ultimately easier to read and less error-prone.

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -0,0 +1,55 @@
...
+    String[] currentRow = {}; // array stores single row of CSV data (one advertisement)
+    // read header rows separately
+    String primaryHeader = bufferedReader.readLine();
+    String secondaryHeader = bufferedReader.readLine();
+    int rowIndex = START_ROW_INDEX; 
+    String row = ""; 
+    while((row = bufferedReader.readLine()) != null && (rowIndex < END_ROW_INDEX)) {
+      /* Regex splits the row into a String array by columns (columns are indicated in 
+      the csv by commas.) It ignores commas inside quotation marks to avoid spliting 
+      individual text fields (the text field "North Carolina, Florida" is not split
```
andrewlathamgoogle, line 33: How does the CSV itself know this?

server/src/main/java/com/google/sps/utils/Ad.java
kirakira0: Ok

server/src/main/java/com/google/sps/utils/Ad.java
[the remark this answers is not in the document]
kirakira0: I see. I've replaced those lines

server/src/main/java/com/google/sps/utils/Ad.java
kirakira0: Got it

server/src/main/java/com/google/sps/utils/Ad.java
kirakira0: Added

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Added constants

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Updated name

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: Added space

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: Added space

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
kirakira0: I receive strings like "100000" that I need to save to the db as longs.

server/src/main/java/com/google/sps/utils/ReportReader.java
kirakira0: It doesn't. I need the regex to only split on the comma if there's an even number of quotes ahead of it.

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,306 @@
+/* Description: Ad defines the ad objects that we write to Firebase.
+ * Author: Kira Toal
+ * Date: June 24, 2020
+ */
```
ailemeo, line 4: Put the Javadoc on top of the class 'Ad' itself.

Nit: empty line between the package and the imports.

server/src/main/java/com/google/sps/utils/Ad.java
```diff
@@ -0,0 +1,306 @@
...
+  public String getContentSentiment() {
+    return this.contentSentiment; 
+  }
+  public String getContentTerms() {
+    return this.contentTerms; 
+  }
+  /*
+   * Builds an instance of Ad. 
+   */ 
+  static class Builder {
```
ailemeo, line 222: Good choice with the Builder pattern.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,116 @@
+/* 
+ * Description: Utility class that processes a CSV row and converts between row format and Ad format. 
+ * Author: Kira Toal
+ * Date: June 24, 2020
+ */
```
ailemeo, line 5: Same as above. Javadoc goes on the class. For the other classes in the PR as well.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public final class AdRowProcessor {
+
+  private static final int ID_ROW = 0;
```
ailemeo, line 21: are these technically rows or columns?

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public final class AdRowProcessor {
```
ailemeo, line 19: Why are there no tests?

kirakira0: As we discussed yesterday, I will be making a new PR for the tests. I am resolving this thread to indicate that it has been addressed.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+        .collect(Collectors.toList());
+    return trimmedList;  
+  }
+
+  public static long convertStringToLong(String str) {
+    try {
+      return (long) Double.parseDouble(str);
+    } catch (IllegalArgumentException e) {
+      System.out.println("Error: " + e);
+      System.exit(0);
```
rob-marcus, line 117: I believe since this you are exiting due to an error, this would qualify as 'abnormal' termination. In that case, you should use `System.exit(1)` or `System.exit(-1)` - `System.exit(0)` indicates normal execution. You could alternatively just `return -1` here instead, which i think you do below but that will never be reached if i understand `System.exit` correctly. 

https://www.baeldung.com/java-system-exit

server/src/main/java/com/google/sps/utils/WriteAd.java
```diff
@@ -0,0 +1,36 @@
...
+
+/* 
+ * Description: Utility class that writes as objects to Firestore. 
+ * Author: Kira Toal
+ * Date: July 6, 2020
+ */ 
+public final class WriteAd {
+
+  public static void writeAd(Ad ad, String collection, String pathToServiceAccount, String databaseURL) throws Exception {
+    // Initialize app.
```
rob-marcus, line 21: Think their should be vertical whitespace between `writeAd` and the comment, but not 100% sure.

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -26,104 +10,112 @@
...
     String[] impressionsArray = formatImpressionField(impressionsField); 
+    
     // If csv field uses "<=", arr has length 1 and min number of impressions is 0.
     if (impressionsArray.length > 1) {
       try {
         return Long.parseLong(impressionsArray[0]);
-      } catch (IllegalArgumentException e) {
-        return -1; 
+      } catch (NumberFormatException e) {
+        System.out.println("Error: " + e);
```
andrewlathamgoogle, line 80: Always throw exceptions, never explicitly exit.  A caller function may want to catch your exceptions and do its own thing with them.  For instance you may write a test to verify that this method does in fact throw the exception, and you'd want to catch the exception and verify that it's correct.  System.exit would prevent that.

You also don't need to System.out.println.  When your program crashes because a thrown exception goes uncaught, it should automatically dump the exception and stack trace.

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -1,55 +1,55 @@
...
-      currentRow = row.split(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", -1); 
-      for (int currentColumn = 0; currentColumn < currentRow.length; currentColumn++) {
-        currentRow[currentColumn] = currentRow[currentColumn].replace("\"", ""); // remove extra quotes
+  public static void readCSV(InputStream csv) throws Exception {
+    CSVParser csvParser = CSVFormat.DEFAULT.parse(new InputStreamReader(csv));
+    List<CSVRecord> csvRecordList = csvParser.getRecords(); 
+    CSVRecord[] csvRecords = new CSVRecord[csvRecordList.size()];
+    csvRecords = csvRecordList.toArray(csvRecords);
+    int currentRowIndex = 0; 
+    while (currentRowIndex <= END_ROW_INDEX + 2) { // Adding two accounts for the two header rows.
```
andrewlathamgoogle, line 37: This +2 business is confusing

server/src/main/java/com/google/sps/utils/WriteAd.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.firebase.cloud.FirestoreClient;
+import java.io.FileInputStream;
+import java.io.InputStream;
+
+/* 
+ * Description: Utility class that writes as objects to Firestore. 
+ * Author: Kira Toal
+ * Date: July 6, 2020
+ */ 
+public final class WriteAd {
```
andrewlathamgoogle, line 18: Name this class something like FirebaseAdWriter.  WriteAd is a verb.

server/src/main/java/com/google/sps/utils/WriteAd.java
andrewlathamgoogle: Not on the first line of the method

server/src/main/java/com/google/sps/utils/AdRowProcessor.java
```diff
@@ -0,0 +1,117 @@
...
+        .map(String::trim)
+        .collect(Collectors.toList());
+    return trimmedList;  
+  }
+
+  public static long convertStringToLong(String str) {
+    try {
+      return (long) Double.parseDouble(str);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException();
```
ailemeo, line 114: This will hide the details/stacktrace from the actual exception `e`. Rethrow the exception, or better still remove the `catch` altogether if it doesn't do anything other than rethrowing. 
Also add `throws IllegalArgumentException` to the method definition.

server/src/main/java/com/google/sps/utils/FirebaseAdWriter.java
```diff
@@ -0,0 +1,36 @@
...
+
+/* 
+ * Description: Utility class that writes as objects to Firestore. 
+ * Author: Kira Toal
+ * Date: July 6, 2020
+ */ 
+public final class FirebaseAdWriter {
+
+  public static void writeAd(Ad ad, String collection, String pathToServiceAccount, String databaseURL) throws Exception {
+    // Initialize app.
```
scarletnguyen13, line 21: Why having this comment here while the init App is actually on line 28?

PR #11, opened 2020-07-13
Refactor alg cloud functions

rob-marcus:
Motivation for this PR: 
- Simplify firebase -> algolia export path and deployment method. 
- Change dependency injection approach from dynamic functions with spying to static exports and stubbing. 
- Refactor file system, make way for additional cloud functions for other query types. 
- Add a `readme.md` to facilitate others adding their own code easily.
- Prepare for dev and prod environments (See `readme.md`)


server/utils/cloudFunctions/functions/algoliaCloudFunctions/devAlgoliaFunctions.js
```diff
@@ -0,0 +1,52 @@
...
+ * Use default value unless mocking an algoliaOperation.
+ * @return {!Promise} 
+ */
+exports.devDeleteRecord = 
+    DEV_ADS_DOCS.onDelete(snapshot => {
+      return deleteRecord(DEV_ADS_INDEX, snapshot);
+    });
+
+
+
```
kirakira0, line 52: nit: extra whitespace

server/utils/cloudFunctions/functions/algoliaCloudFunctions/prodAlgoliaFunctions.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Description: Cloud functions for PROD firebase->algolia features. 
+ *              Cloud functions must be created as static JS objects
+ *              hence not everything can be DRY. 
+ *              Three basic operations supported: create, update, delete.
+ *              Vocab: a firestore collection = algolia index
+ *                     an ad entity = algolia record
```
kirakira0, line 7: To keep consistent with Firestore vocab, use "ad document" as opposed to "ad entity"

server/utils/cloudFunctions/functions/firebaseConfig.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Description: Initialize the index environment
+ *              Basically prepare firestore and algolia
```
kirakira0, line 3: nit: Lots of "basically"s in your comments. I think it's a little unsure-sounding/informal

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaFunctions.test.js
```diff
@@ -0,0 +1,143 @@
...
+ * Date: July 11, 2020
+ */
+
+// Import the testing environment configuration
+const { test, assert, sinon } = require('../testConfig');
+
+// Import the generic cloud functions.
+const devAlgoliaFunctions = 
+    require('../../algoliaCloudFunctions/devAlgoliaFunctions');
+const prodAlgoliaFunctions =
```
andrewlathamgoogle, line 16: Why do you have prod stuff in here?  Unless there's a good reason for it, it's incredibly dangerous to be messing with prod in your tests.  What if you wrote the test wrong and end up actually writing to prod?  What if some future engineer comes along and tinkers with the test and accidentally writes to prod?

Is there something you need to test with prod that you can't test with devel?

rob-marcus: Good points, had not considered the danger of prod. First instinct was just to have a test for it, but I suppose since it is supposed to just be a mirror of prod, the onus is on the author to ensure synchronicity, not the tests.

server/utils/cloudFunctions/functions/firebaseConfig.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Description: Initialize the index environment by 
+ *              initializng the firebase api to access the dev/prod ads db
+ * Author: Robert Marcus
+ * Date: July 7, 2020
+ */
+
+// Initialize the firebase functions sdk. 
+const functions = require('firebase-functions');
```
scarletnguyen13, line 9: nit: please make this more clear - something like `firebaseFunctions` would make more sense, so you don't need to go back to this file to see what `functions` actually mean

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaFunctions.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+// Import mock saveObject, mock deleteObject, and helpers.
+const { mockDeleteObject, mockSaveObject } = require('./algoliaMocks');
+
+// Restore the stubs after each test. 
+afterEach(() => {
+  sinon.restore();
+});
+
+describe("test_devCreateRecord", () => {
```
ailemeo, line 28: I think you can remove "dev" from these test descriptions. So just test_CreateRecord for example.
It is expected that unit tests run on the dev not prod.

client/build/static/js/2.d0237a51.chunk.js
```diff
@@ -1,3 +0,0 @@
-/*! For license information please see 2.d0237a51.chunk.js.LICENSE.txt */
```
ailemeo, line 1: What do/did these files have to do with algolia?

rob-marcus: Nothing. Made the mistake of compiling the web app off this branch I believe, it created a bunch of cached files that I accidentally added to the pr, then I removed them. Incidental garbage clean up basically.

PR #14, opened 2020-07-15
Functional Firebase Ad Reader

kirakira0:
Small PR demonstrating a basic server-side Firestore query.


server/src/main/java/com/google/sps/utils/FirebaseAdReader.java
```diff
@@ -0,0 +1,55 @@
...
+ * Description: Utility class that reads ad objects from Firestore. 
+ * Author: Kira Toal
+ * Date: July 14, 2020
+ */ 
+public class FirebaseAdReader {
+
+  private static final String COLLECTION = "testing";
+  private static final String DATABASE_URL = "https://step9-2020-capstone.firebaseio.com"; 
+  private static final String PATH_TO_SERVICE_ACCOUNT = "./serviceAccountKey.json"; 
+  private static final int TIMEOUT_ALLOWANCE = 30; // Time in seconds before future.get() times out.
```
rob-marcus, line 32: Is this number experimental?

kirakira0: At the moment it's an estimate based on how many ads are in the collection I'm using. I'm trying to figure out a good way to dynamically adjust how much time I allow before a timeout.

PR #15, opened 2020-07-15
Count advertisers cloud functions

rob-marcus:
Don't worry, it's not actually 800 loc. Please read `readme.md` before reviewing in case it may answer any comments. 

**The test file may not load by default because it is too large, you may have to click `load diff`**

Breakdown, +-: 
- 50-100 lines of config, readme.
- 250 lines of cloud function code.
- 400 lines of tests.
- I added a `describe` block to the `algolia` tests, GitHub thinks I rewrote a ton of lines of that file. It's identical, just slightly indented. 

Motivation for this pull: 
- Make client side query for my chart a lot easier to write and test. 
- Make the general query (`return top 15 advertisers 2018`) fast and affordable. 
  - It looks like firebase only caches for ["as long as there is an active listener for that data"](https://stackoverflow.com/questions/38423277/does-firebase-cache-the-data)
  - i.e., don't expect much caching by default and/or without a lot of effort. 
  - ~$0.40-$0.50 per 75K reads, we _*do not*_ want to read the whole db frequently. 
  - `ahead of time` approach has immediate server side cost hit, but that's a lot cheaper on all accounts than a `just-in-time` client-side query approach. 

All tests passing. Code coverage is at 97.5%. The one LOC not tested is a `catch` block that I couldn't trigger, will only practically trigger if firestore dies during a read or a write. 

Screen Shot 2020-07-18 at 6 26 57 PM


server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/devCountAdvertisersFunctions.js
```diff
@@ -0,0 +1,38 @@
...
+// Import links to firebase dev advertisers collections.
+// These are used to write directly to a collection.
+const { DEV_AGGREGATES_COLLECTION } = require('./countAdvertisersConfig');
+
+// Import helper functions.
+const { incrementAdvertiserCount, decrementAdvertiserCount } = 
+    require('./countAdvertisersHelpers');
+
+/**
+ * Increments an advertisers annual aggregate by one for an ad in a given year
```
kirakira0, line 19: You miss a lot of apostrophes w/ the word "advertisers". I know the grammar of the comments isn't the most "important" part of the PR, but I do think it's worth the fix.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaFunctions.test.js
```diff
@@ -20,63 +20,66 @@ const { DEV_ADS_INDEX } =
...
   });
-});
 
-describe("test_deleteRecord", () => {
-  it('should call deleteObject with correct values, return its output', () => {
-    const deleteStub = 
-        sinon.stub(DEV_ADS_INDEX, "deleteObject").callsFake(mockDeleteObject);
-    const snap = firestoreMock.exampleDocumentSnapshot();
+  describe("test_deleteRecord", () => {
+    it('should call deleteObject with correct values, return its output', () => {
```
kirakira0, line 70: Do you test for create, save, or deleteRecord getting called with incorrect values?

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersConfig.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Description: initialize the environment and firestore db access.
+ *              `aggregates` refers to a collection of aggregated ad data.
```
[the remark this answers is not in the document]
rob-marcus: Yes. I have made a `readme.md` to address this comment and some others to clarify the purpose and intention.

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
```diff
@@ -0,0 +1,94 @@
...
+                .collection("advertisers")
+                .doc(advertiser);
+
+  return advertiserCountReference;
+}
+
+/**
+ * Decrements an advertisers annual aggregate by one for an ad in a given year
+ * in dev env. Annual aggregate stored in firebase collection `DEV_AGGREGATES_COLLECTION`.
+ * Years possible are 2018, 2019, 2020. Event triggered on delete.
```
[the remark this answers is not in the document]
rob-marcus: Oops. This is not a strict limit, carry over from earlier implementation. See `readme.md` for more info.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaFunctions.test.js
rob-marcus: Not currently. Since these functions are in practice firewall'd to only use ads created by your `adWriter.java`, I am presuming that the ads are well-formed/typed.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
```diff
@@ -0,0 +1,406 @@
...
+
+// Save some space for test explanation in the test description
+const SHOULD_INCREMENT = "should increment advertisers ad count "; 
+
+describe('Count advertisers cloud functions', () => {
+  // After every 'describe' block, reset the test environments. 
+  // Since these are live dev db's, this process can be very flaky.
+  // In general, don't expect an 'advertiser' will be reset from 
+  // test to test. 
+  after(() => {
```
scarletnguyen13, line 60: this test can be shorter by having an array of collections `const collections = ['dev_ads', ''dev_aggregates/2018/advertisers', 'dev_aggregates/2019/advertisers', 'dev_aggregates/2020/advertisers'];` then map over the array and `deleteCollection` :D

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
```diff
@@ -0,0 +1,406 @@
...
+              .collection("advertisers")
+              .doc(advertiser).get().then(function(doc) {
+                return assert.equal(doc.data().numberOfAds, 1);
+              }).catch(err => console.log(err));
+
+          return;
+        });
+      }); 
+    });   
+
```
scarletnguyen13, line 202: Do we have tests for when it SHOULD NOT increment/decrement?

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
```diff
@@ -0,0 +1,406 @@
...
+              .doc("adv_g").get().then(function(doc) {
+                return assert.equal(doc.data().numberOfAds, 1);
+              }).catch(err => console.log(err));
+          
+          return;
+        });
+      }); 
+    });
+
+    it("should fail with malformed inputs", () => {
```
scarletnguyen13, line 234: Test inputs with correct form but not existed - i.e. 2020-02-30

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/devCountAdvertisersFunctions.js
```diff
@@ -0,0 +1,47 @@
...
+const { incrementAdvertiserCount, decrementAdvertiserCount } = 
+    require('./countAdvertisersHelpers');
+
+/**
+ * Increments an advertiser's annual ad count by one for an ad in a given year.
+ * in dev env. Annual ad count stored in firebase collection `DEV_AGGREGATES_COLLECTION`.
+ * Event triggered on create.
+ * @returns {!Promise}     
+ */
+exports.devCountAdvertisersOnCreate =
```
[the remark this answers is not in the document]
rob-marcus: Default way to export cloud functions.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
rob-marcus: I'd consider the tests verifying only the correct year is incremented as a sort of `SHOULD NOT increment an ad count in the wrong year` type of test. Otherwise it's always correct to increment. 

Re. SHOULD NOT decrement, yes, test `should only decrement if ad count is positive` verifies when it should not decrement.

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
```diff
@@ -0,0 +1,118 @@
...
+ * Decrements an advertiser's annual ad count by one for an ad in a given year.
+ * Event triggered on delete. If the advertiser's count document does not 
+ * exist in the relevant collection, a 500 error will be thrown.
+ * If `advertiser` or `startDate` field is not well-typed/formed, 
+ * an assertion error will be thrown. 
+ * @param {Object} snapshot a json string
+ * @param {Object} collection reference to a particular firestore collection
+ * @returns {!Promise} 
+ */
+function decrementAdvertiserCount(snapshot, collection) {
```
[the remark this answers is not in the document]
rob-marcus: I agree, it's not 100% dry. However, I think it's an acceptably low level of repetition - I think any single helper function that would replicate the behavior would be really verbose and overall not an improvement to code structure or readability.

server/utils/cloudFunctions/functions/test/deleteCollection.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Description: code copied from here: `https://firebase.google.com/docs/firestore/manage-data/delete-data`
+ *              because firestore doesn't have a built in way to delete/reset
+ *              a (sub)collection. 
+ */
+
+const DEFAULT_BATCH_SIZE = 10;
+
+async function deleteCollection(
```
andrewlathamgoogle, line 9: Why is this in the test directory?

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
```diff
@@ -0,0 +1,121 @@
...
+
+      // Only decrement if the numberOfAds is positive.
+      const change = doc.data().numberOfAds > 0 ? -1 : 0;
+
+      return advertiserCountReference
+                .update({numberOfAds: firestoreFieldValue.increment(change)});
+    } else {
+      
+      // If the document doesn't exist throw an error.
+      const errorMessage = { code: 500,
```
andrewlathamgoogle, line 83: What do you intend to do when this occurs?  Manually update the data?

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+function decrementAdvertiserCount(snapshot, collection) {
+  const advertiserCountReference = 
+      getAdvertiserCountReference(snapshot, collection);
+
+  return advertiserCountReference.get().then(function(doc) {
+    if (doc.exists) {
+
+      // Only decrement if the numberOfAds is positive.
+      const change = doc.data().numberOfAds > 0 ? -1 : 0;
```
andrewlathamgoogle, line 76: Why would we ever decrement when the advertiser already believes it has zero ads?  This would indicate that your data is already in an invalid state and you should at least log a warning.

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
```diff
@@ -0,0 +1,121 @@
...
+ * @param {Object} collection reference to a particular firestore collection
+ * @returns {Object} 
+ */
+function getAdvertiserCountReference(snapshot, collection) {
+  const data = snapshot.data();
+  const advertiser = data.advertiser;
+  const startDate = data.startDate;
+  const isoStartDate = new Date(startDate);
+
+  checkAdvertiser(advertiser);
```
andrewlathamgoogle, line 45: What is the purpose of these two methods?  If all you're doing is reading, you shouldn't have to validate your output -- validation is done at write time to ensure that the DB is always in a valid state.

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
```diff
@@ -0,0 +1,121 @@
+/**
+ * Description: Helper functions for count advertisers cloud functions.
+ * Author: Robert Marcus
+ * Date: July 14, 2020
+ */
+
+const { assert } = require('../test/testConfig');
```
andrewlathamgoogle, line 7: This isn't a test so you shouldn't be using assert.  If there's an error, throw an exception.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
```diff
@@ -0,0 +1,400 @@
...
+      const snapOne = 
+          snapFromJson({advertiser, startDate: "2018-10-15"}, DEV_ADS_PATH);
+      const snapTwo = 
+          snapFromJson({advertiser, startDate: "2019-10-15"}, DEV_ADS_PATH);
+      const snapThree = 
+          snapFromJson({advertiser, startDate: "2020-10-15"}, DEV_ADS_PATH);
+
+      const wrapper = firestoreWrap(devCountAdvertisersOnCreate);
+
+      return wrapper(snapOne).then(() => {
```
andrewlathamgoogle, line 95: You should be able to combine all of these in one statement without this awkward nesting.

server/utils/cloudFunctions/functions/test/deleteCollection.js
rob-marcus: It seemed appropriate to put it there because the only legitimate use case I could think of for it was for tearing down collections before/after tests.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
```diff
@@ -0,0 +1,400 @@
...
+              .doc(advertiser).get().then(function(doc) {
+                return assert.equal(doc.data().numberOfAds, 1);
+              }).catch(err => console.log(err));
+
+          return;
+        });
+      }); 
+    });   
+
+    it(SHOULD_INCREMENT + 'only for an advertisers own ads', () => {
```
andrewlathamgoogle, line 201: this description is kinda unclear, it's really testing two different advertisers.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
```diff
@@ -0,0 +1,400 @@
...
+      const snapThree = 
+          snapFromJson({advertiser: "adv", startDate: "201-910-15"}, 
+                        DEV_ADS_PATH);
+
+      const wrapper = firestoreWrap(devCountAdvertisersOnCreate);
+
+      assert.throws(() => wrapper(snapOne), Error);
+      assert.throws(() => wrapper(snapTwo), Error);
+      assert.throws(() => wrapper(snapThree), Error);
+      assert.throws(() => wrapper(snapFour), Error);
```
andrewlathamgoogle, line 247: snapFour?

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
rob-marcus: It will be helpful with firebase redeploy, as I believe you can blacklist certain failures by error code to not re-attempt them. Basically yes, it'll make life easier if we need to go chase through the logs.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
```diff
@@ -0,0 +1,400 @@
...
+    });  
+  });
+  
+  describe("test_countAdvertisersOnDelete", () => {
+    it("should decrement ad count when deleting an advertisers ad", () => {
+      // The 'after' should cleanup the DB of "adv_a" references and values
+      // in time, but it can be flaky. 
+      // Easy way to avoid this problem is just to switch to adv_A, adv_B, 
+      // and so on. 
+      const snap = snapFromJson({advertiser: "adv_A", startDate: "2018-10-15"},
```
andrewlathamgoogle, line 257: be consistent in how you write these tests.  In the others you declare advertiser separately rather than inline.

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
rob-marcus: Will rewrite this to include a log.

server/utils/cloudFunctions/functions/countAdvertisersCloudFunctions/countAdvertisersHelpers.js
rob-marcus: Will leave validation to write time only.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
rob-marcus: Dead code. oops.

server/utils/cloudFunctions/functions/test/countAdvertisersTests/countAdvertisersFunctions.test.js
rob-marcus: Done where possible with `Promise.allSettled`.

PR #16, opened 2020-07-16
Fix eslint errors

scarletnguyen13:


client/src/components/Visualization/charts/Geochart.js
```diff
@@ -1,43 +1,50 @@
...
 
   return (
     <div className="search-header center">
       <p>Impressions Geochart</p>
-      <Chart chartType="GeoChart" width="700px" height="400px" data={getData()} options={options} />
+      <Chart
+        chartType="GeoChart"
+        width="700px"
+        height="400px"
+        data={getData()}
```
rob-marcus, line 43: Think this should be `{ getData() }`

scarletnguyen13: ah for field variables we don't need to apply that rule :D

client/src/components/Visualization/charts/Geochart.js
```diff
@@ -1,43 +1,50 @@
...
+    region: 'US',
+    tooltip: { trigger: 'focus' }, // Trigger info box on mouse hover over state.
   };
 
   return (
     <div className="search-header center">
       <p>Impressions Geochart</p>
-      <Chart chartType="GeoChart" width="700px" height="400px" data={getData()} options={options} />
+      <Chart
+        chartType="GeoChart"
```
andrewlathamgoogle, line 40: I'm surprised lint complained about silly things like spacing in your single-line JSON objects but not your mixing of single and double quotes in the same file.

PR #17, opened 2020-07-16
Add Analysis UI component in react

scarletnguyen13:
Design:

Screen Shot 2020-07-16 at 8 29 54 PM


client/src/App.js
```diff
@@ -2,22 +2,16 @@ import './App.css';
 
 import React, { useState } from 'react';
 
+import { PAGE_MAP } from './constants/header_constant';
```
rob-marcus, line 5: Minor nit: the filename `header_constants` would probably be better because I see you have more than one constant exported.

client/src/components/SentimentAnalysis/Helpers.js
```diff
@@ -0,0 +1,95 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const HEADLINE_PLACEHOLDER =
+  'i.e. Fighting Climate Change | Pete Buttigieg 2020 | Learn More';
```
rob-marcus, line 5: Minor nit: `i.e.` means that is, you want `e.g.`, i.e., for example :).

client/src/components/SentimentAnalysis/Helpers.js
```diff
@@ -0,0 +1,95 @@
...
+const CONTENT_PLACEHOLDER =
+  "i.e. Climate change is a life and death issue for our generation. It's time for bold action.";
+
+const INPUT_LIST = [
+  { label: 'HEADER', placeholder: HEADLINE_PLACEHOLDER },
+  { label: 'CONTENT', placeholder: CONTENT_PLACEHOLDER },
+];
+
+const AnalysisInput = props => {
```
rob-marcus, line 14: For the future: 
- I'm not sure if this is searching through ads in our db and just using the lang analysis results already computed or calling the api and recomputing (haven't read all the way through yet.)
  - If you are doing the latter, consider looking into adding captcha + a 1000 character limit (that's how big one unit of an api call is for the google api.)

client/src/components/SentimentAnalysis/Helpers.js
```diff
@@ -0,0 +1,95 @@
...
+  const { label, placeholder } = props;
+
+  return (
+    <div className="search-header center">
+      <h3 className="analysis-label">{label}</h3>
+      <textarea
+        type="text"
+        rows={3}
```
rob-marcus, line 22: magic number comment please.

client/src/components/SentimentAnalysis/Helpers.js
```diff
@@ -0,0 +1,95 @@
...
+AnalysisInput.propTypes = {
+  label: PropTypes.string,
+  placeholder: PropTypes.string,
+};
+
+const ColorBar = props => {
+  const { score, magnitude } = props;
+
+  const colorMap = {
+    negative: { label: 'Negative', color: 'red', background: '#ffbbc6' },
```
rob-marcus, line 39: colors.css

client/src/components/SentimentAnalysis/Helpers.js
scarletnguyen13: Really good point, I'm doing the latter one and this comment is really useful to me, thank you! :)

client/src/components/SentimentAnalysis/Helpers.js
```diff
@@ -0,0 +1,81 @@
...
+const TermsDisplay = props => {
+  const { termList } = props;
+
+  return (
+    <div className="analysis-text-container">
+      {termList.map((termString, index) => {
+        const termObject = JSON.parse(termString);
+        return (
+          <div key={index} className="center">
+            <div className="analysis-text" id="special-text">
```
andrewlathamgoogle, line 65: It seems strange to me to have a helper that has an ID in it.  IDs should be unique, meaning this helper would only ever be used once and you might as well just inline it.

client/src/components/SentimentAnalysis/SentimentAnalysis.js
```diff
@@ -0,0 +1,72 @@
...
+        {isVerified ? (
+          <button className="analysis-button">Analyze</button>
+        ) : (
+          <ReCAPTCHA sitekey={CLIENT_KEY} onChange={onChange} />
+        )}
+      </div>
+      <div className="analysis-result-container center">
+        <div className="analysis-card center">
+          <h4>Header Analysis</h4>
+          <ColorBar score={0.4} magnitude={0.7} />
```
andrewlathamgoogle, line 59: Are 'score' and 'magnitude' here part of your sample data?  If so make them constants and put them up with the other ones so it's clear what is mock and what isn't.

client/src/constants/analysis_constants.js
```diff
@@ -0,0 +1,23 @@
+const HEADLINE_PLACEHOLDER =
```
andrewlathamgoogle, line 1: Put all your mock/placeholder stuff in one place please.  If you scatter it all over your code there's a risk you'll forget one (or whoever has to come in and change this code after you will miss one).

client/src/firebase/FirestoreDocumentReader.js
```diff
@@ -7,8 +7,8 @@
```
andrewlathamgoogle, line 7: For future PRs can you please separate lint stuff and relevant stuff.  This PR looks a lot bigger than it is because it has a lot of unrelated lint changes.

scarletnguyen13: Yessir 👍

PR #18, opened 2020-07-16
Algolia integration tests

rob-marcus:
Integration tests for firebase -> algolia. 

P.S. in case you don't read it in the doc, these tests take +- 10s because that's the maximum amount of latency for firestore to execute a function. 

After integrating `try/catch` repeat logic: 
Screen Shot 2020-07-22 at 2 22 08 PM
As much as 4x faster than before with naive `timeout` approach 🎉 !!!!!


server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,218 @@
...
+        }, TIMEOUT_10S);
+      });
+    }).timeout(TIMEOUT_15S);
+    
+    it("should propogate multiple ad documents without overwriting", function(done) {
+      const adList = [{advertiser: "b", startDate: "2019-10-15"},
+                      {advertiser: "c", startDate: "2019-10-15"},
+                      {advertiser: "d", startDate: "2019-10-16"}];
+
+      DEV_ADS_COLLECTION.doc("b").set(adList[0]).then(() => {
```
scarletnguyen13, line 96: Having callback hell like this is not ideal at all, is there other way we can get away with this (i.e. using Await?)

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,218 @@
...
+      });
+    }).timeout(TIMEOUT_15S); 
+
+    it("should delete the right algolia record when a firestore document is deleted", function(done) {
+      const adOne = {advertiser: "i", startDate: "2019-10-15"};
+      const adTwo = {advertiser: "j", startDate: "2019-10-15"};
+
+      DEV_ADS_COLLECTION.doc("i").set(adOne).then(() => {
+        DEV_ADS_COLLECTION.doc("j").set(adTwo).then(() => {
+          DEV_ADS_COLLECTION.doc("i").delete().then(() => {
```
scarletnguyen13, line 195: if you're using alphabetical letters for testing, what happen if you ran out of letters? :)

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
rob-marcus: I agree, this is a rats nest of callback nonsense. I have experimented with using  await and such, and I've also experimented with using stuff like `Promise.all`. 
- The main problem I had with using `await` was significantly more test flakiness. 
- I had weird issues using `Promise.all` with `chai` where assertions wouldn't fail properly. 

EOD, promise nesting makes two things very easy: keeps things in scope easily, makes sure things get executed in the proper order only after completion. It's also a lot less flaky than using await. Since this is just for integration testing and not code that will need to change much ever, I think it's an acceptable compromise for the ugliness.  

Going to leave this comment `unresolved` for visibility in case others have a similar q.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
rob-marcus: `.doc("z"), .doc("aa"), .doc("ab"),...`. I see your point, but I think alphabetical enumeration is the best worst option for easy indexing.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,237 @@
+/**
+ * Description: Integration tests for algolia cloud functions.
+ *              Can be flaky. If an error like `ObjectID` does not exist, 
+ *              it is probably a problem with algolia flaking. When in doubt,
+ *              re-run the tests. Compile with `npm run test`.
+ * Notes:
+ * - For simplicity, advertisment json's are denoted as `ad`, and
```
kirakira0, line 7: JSONs, not json's (since it's plural, not possessive or "JSON is")

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,237 @@
...
+ * Notes:
+ * - For simplicity, advertisment json's are denoted as `ad`, and 
+ *   the advertiser is used as the documents `primary key`. 
+ *   - In practice, primary key is a random string.
+ *   - If used something like `adOne` as `primary key`, would probably
+ *     run into a lot of collisions with other `adOne`'s not deleted
+ *     fast enough between tests (see note about `after` cleanup below.)
+ *   - Advertiser naming convention for this test suite is alphabetical 
+ *     enumeration. This is in contrast to the `countAdvertisersFunction`
+ *     test suite, which used alphabetical `ad_x` enumeration.
```
kirakira0, line 15: Are lengthy notes like this supposed to go in the file? I'm not sure of the convention but it seems like it might be better placed in a different file. Not sure though, maybe something to check w/ the hosts.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,237 @@
...
+ * how errors are handled/resolved.   
+ * Author: Robert Marcus
+ * Date: July 16, 2020
+ */
+
+
+// Import the testing environment configuration.
+const { chai,
+        TIMEOUT_10S,
+        TIMEOUT_15S,
```
kirakira0, line 75: Why 10 and 15 seconds? Will this ever change?

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,237 @@
...
+// Import firebase environment configuration.
+const { DB } = require('../../firebaseConfig');
+
+// Import algolia dev environment configuration.
+const { DEV_ADS_INDEX } = 
+    require('../../algoliaCloudFunctions/algoliaConfig');
+
+describe("Algolia integrations tests", () => {
+  // After every 'describe' block, reset the test environments. 
+  // Since these are live dev db's, this process can be very flaky.
```
kirakira0, line 89: Plural, not possessive -- "dev DBs", no commas

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
rob-marcus: There is a comment about this above and where these constants are declared, but I added another comment here so one doesn't have to search for context.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,242 @@
+/**
+ * Description: Integration tests for algolia cloud functions.
+ *              Can be flaky. If an error like `ObjectID` does not exist,
```
andrewlathamgoogle, line 3: Is it the tests that are flaky or Algolia itself?  Don't submit flaky tests.  You will hate yourself for it later when you want to make a quick change to these tests and never know whether the failure is due to your new test or due to flakiness.

Plus in a production engineering environment flaky tests are a high-priority catch-and-kill task.  They drive engineers absolutely up the wall.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,242 @@
+/**
+ * Description: Integration tests for algolia cloud functions.
+ *              Can be flaky. If an error like `ObjectID` does not exist, 
+ *              it is probably a problem with algolia flaking. When in doubt,
+ *              re-run the tests. Compile with `npm run test`.
+ * Notes:
+ * - For simplicity, advertisment JSONs are denoted as `ad`, and
```
andrewlathamgoogle, line 7: *advertisement

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,242 @@
...
+ *              it is probably a problem with algolia flaking. When in doubt,
+ *              re-run the tests. Compile with `npm run test`.
+ * Notes:
+ * - For simplicity, advertisment JSONs are denoted as `ad`, and 
+ *   the advertiser is used as the documents `primary key`. 
+ *   - In practice, primary key is a random string.
+ *   - If used something like `adOne` as `primary key`, would probably
+ *     run into a lot of collisions with other `adOne`'s not deleted
+ *     fast enough between tests (see note about `after` cleanup below.)
+ *   - Advertiser naming convention for this test suite is alphabetical
```
andrewlathamgoogle, line 13: Can you just read the system clock and use that to name your entities to ensure they're always unique

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,242 @@
...
+// Import firebase environment configuration.
+const { DB } = require('../../firebaseConfig');
+
+// Import algolia dev environment configuration.
+const { DEV_ADS_INDEX } = 
+    require('../../algoliaCloudFunctions/algoliaConfig');
+
+describe("Algolia integrations tests", () => {
+  // After every 'describe' block, reset the test environments. 
+  // Since these are live dev DBs, this process can be very flaky.
```
andrewlathamgoogle, line 94: If deletion is flaky, tests should never be dependent on a previous deletion being successful.

Make each test case unique, delete on best-effort just to conserve space, and in practice you would set up an alert to yell at you when the DB got too big so you could go wipe it manually.

As I said earlier you should be able to achieve this by checking the system clock.  If that doesn't work you could also use RNG to assign a random value to insert at the start of every name.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -0,0 +1,242 @@
...
+        });
+      });
+    }).timeout(TIMEOUT_15S);
+
+    it("should propogate a firestore document update to " + 
+       "the proper algolia record", function(done) {
+      const adOne = {advertiser: "f", startDate: "2019-10-15"};
+      const adTwo = {advertiser: "g", startDate: "2019-10-15"};
+    
+      Promise.allSettled([DEV_ADS_COLLECTION.doc("f").set(adOne),
```
andrewlathamgoogle, line 165: This promise chaining gets hard to read.  My eyes keep having to bounce around vertically to align your blocks.   Depending on your stack dumps it may also be harder to trace.
Is it possible to rewrite it as something like:

Promise writeF = DEV_ADS_COLLECTION.doc("f").set(adOne);
Promise writeG = DEV_ADS_COLLECTION.doc("g").set(adOne);
await Promise.allSettled(writeF, writeG);
await DEV_ADS_COLLECTION.doc("f").update({startDate: "2020-10-15"}).timeout(TIMEOUT_15S);
setTimeout(...)

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
rob-marcus: Algolia is flaky, the tests are fine. To clarify the flakiness, I'm experiencing 1 sub test failing for every 10 runs due to algolia flakiness. I will update the JSdocs to de-emphasize this.

server/utils/cloudFunctions/functions/test/algoliaTests/algoliaIntegrationTests.test.js
```diff
@@ -89,154 +53,148 @@ const { DB } = require('../../firebaseConfig');
...
-        setTimeout(function(){
-          DEV_ADS_INDEX.getObject("a").then(content => {
-            chai.expect(content.data).to.deep.equal(ad);
-          }).catch(err => console.log(err));
-
-          done();
-        }, TIMEOUT_10S);
-      });
+    it("should propogate an ad document into algolia index", async () => {
+      const ad = {advertiser: makeRandomID(), startDate: "2019-10-15"};
```
andrewlathamgoogle, line 70: define this as its own variable, to avoid the possibility of ad being modified in one of your methods, thus changing this ID.

server/utils/cloudFunctions/functions/test/promiseTimeout.js
```diff
@@ -0,0 +1,15 @@
...
+ * Description: helper function that returns a promise that will resolve
+ *              after an alloted amount of time.
+ * Date: 7/21/2020
+ * Author: Robbie Marcus
+ */
+
+/**
+ * @param {integer} ms how many ms the promise should wait before resolving.
+ */
+function promiseTimeout(ms) {
```
andrewlathamgoogle, line 11: Why not just put this in retryAssertions instead of its own file

server/utils/cloudFunctions/functions/test/retryAssertions.js
```diff
@@ -0,0 +1,32 @@
...
+ * Wraps an async function/`=> const` with recursive try and catch logic.
+ * If `timeout` grows larger than `TIMEOUT_MAX`, the test will fail
+ * and the error message will be returned as the reason for test failure.
+ * @param {function} assertions some async (lambda/big arrow) function  
+ * @param {integer} timeout number of ms to wait before retrying the assertions
+ */
+const retryAssertions = async function(assertions, timeout) {
+  try {
+    await assertions();
+  } catch (err) {
```
andrewlathamgoogle, line 22: You should check the type of the error and only retry on the algolia error.  Otherwise if your test legitimately fails, it's going to throw an error, and you're going to keep retrying until 30 seconds have been wasted repeatedly retrying the test.

rob-marcus: I tried to implement this but it is proving difficult - a surprising amount of variation in the valid errors going by during the retry process. Will save this for a future PR.

PR #19, opened 2020-07-17
Setup DEV and PROD config for Algolia

scarletnguyen13:


client/src/constants/algolia_config.js
```diff
@@ -1,11 +1,14 @@
 import algoliasearch from 'algoliasearch/lite';
 
-const SEARCH_CLIENT = algoliasearch(
+const searchClient = algoliasearch(
   'E82P8D3AVW',
```
rob-marcus, line 4: I know what these are, but please provide an inline comment that the first is `/* app id = */` and the second is `/* search only api key = */`, or something like that for clarity.

PR #25, opened 2020-07-21
Common advertisers scatter plot

rob-marcus:
In practice, will be top 10 or more. Working off a small `dev_ads` subset. 
Screen Shot 2020-07-21 at 4 53 47 PM
Screen Shot 2020-07-21 at 4 57 59 PM

### DIFF
--- a/client/src/components/Visualization/charts/ScatterPlot.js
+++ b/client/src/components/Visualization/charts/ScatterPlot.js
@@ -1,11 +1,16 @@
-import { VictoryTheme, VictoryScatter, VictoryChart, VictoryAxis, VictoryTooltip, VictoryLabel } from 'victory';
+import { VictoryTheme, 
+         VictoryScatter, 
+         VictoryChart, 
+         VictoryAxis, 
+         VictoryTooltip, 
+         VictoryLabel } from 'victory';
 import { database } from '../../../firebase/firebase';
 import React, { useState, useEffect } from 'react';
 
 /**
  * Creates a hexidecimal color string from a string input. 
  * Taken from here: `https://stackoverflow.com/a/16348977`.
- * Loosely cleaned up for style.
+ * Loosely cleaned up for style and comments.
  * This *should* return a valid hash, but if it doesn't,
  * the browser will just display `#000`.
  * @param {string} str input string 
@@ -39,10 +44,13 @@ function stringToColor(str) {
 function formatAdvertiserCountSnapshot(snap, year) {
   const numberOfAds = snap.data().numberOfAds;
   const advertiser = snap.id;
-  const victoryFormattedAdvertiserCount = {x: parseInt(year), 
-                                           y: parseInt(numberOfAds), 
-                                           fill: stringToColor(advertiser),
-                                           label: `${advertiser} had ${numberOfAds} ads in ${year}.`};
+
+  const victoryFormattedAdvertiserCount = 
+      {x: parseInt(year), 
+       y: parseInt(numberOfAds), 
+       fill: stringToColor(advertiser),
+       label: `${advertiser} had ${numberOfAds} ads in ${year}.`};
+
   return victoryFormattedAdvertiserCount;
 }
 
@@ -74,7 +82,8 @@ function useAdvertisers(year, limit) {
             .collection("advertisers")
             .orderBy("numberOfAds", "desc")
             .limit(limit)
-            .get().then((snapshots) => {
+            .get()
+            .then((snapshots) => {
       const newAdvertisers = snapshots.docs.map(snap => {
         return formatAdvertiserCountSnapshot(snap, year);
       });
--- a/client/src/components/Visualization/charts/ScatterPlot.js
+++ b/client/src/components/Visualization/charts/ScatterPlot.js
@@ -59,9 +59,13 @@ function getChartRange(victoryJsonList) {
 }
 
 /**
- * 
+ * Custom react webhook. Currently calls `dev_aggregates` collection
+ * and queries for the top `year` 
+ * @param {string} year year to query
+ * @param {int} limit maximum number of database documents to return
+ * @return {object}
  */
-function useAdvertisers(year) {
+function useAdvertisers(year, limit) {
   const [advertisers, setAdvertisers] = useState([]);
 
   useEffect(() => {
@@ -69,7 +73,7 @@ function useAdvertisers(year) {
             .doc(year)
             .collection("advertisers")
             .orderBy("numberOfAds", "desc")
-            .limit(4)
+            .limit(limit)
             .get().then((snapshots) => {
       const newAdvertisers = snapshots.docs.map(snap => {
         return formatAdvertiserCountSnapshot(snap, year);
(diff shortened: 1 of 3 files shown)


client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Creates a hexidecimal color string from a string input. 
+ * Taken from here: `https://stackoverflow.com/a/16348977`.
+ * Loosely cleaned up for style and comments.
+ * This *should* return a valid hash, but if it doesn't,
+ * the browser will just display `#000`.
+ * @param {string} str input string 
+ * @return {string}
+ */
+function stringToColor(str) {
```
kirakira0, line 28: Since this is just an arbitrary string, why aren't you just generating a random hex code? Scarlet has a method to do that in utils

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+
+  return victoryFormattedAdvertiserCount;
+}
+
+/**
+ * Returns the exclusive bounds for the range of all data points being charted.
+ * Uses the spread `...` operator.
+ * `https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax` 
+ * @param {List[object]} victoryJsonList a list of objects in a victory-ready 
+ *     format.
```
kirakira0, line 71: nit: I think this can be one line

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+  const advertisers2018 = useAdvertisers("2018", queryLimit);
+  const advertisers2019 = useAdvertisers("2019", queryLimit);
+  const advertisers2020 = useAdvertisers("2020", queryLimit);
+  
+  const advertisers = [...advertisers2018,
+                       ...advertisers2019,
+                       ...advertisers2020];
+
+  const range = getChartRange(advertisers);
+  const chartTitle = `T${queryLimit} Most prolific ad words advertisers/year`;
```
kirakira0, line 121: I thought this was for most prolific **advertisers** not ad words?

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+        tickValues = { [range.min, range.max] }
+        label="# of ads (log scale)"
+      />
+
+      <VictoryScatter
+        labelComponent={ <VictoryTooltip/> }
+        labels={({ datum }) => datum.y}
+        style={{ data: { fill: ({ datum }) => datum.fill }}}
+        size={ 10 }
+        data={ advertisers }
```
kirakira0, line 149: nit: I don't think you need the spaces for variables/values
i.e. size={10} and data={advertisers}

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+      const newAdvertisers = snapshots.docs.map(snap => {
+        return formatAdvertiserCountSnapshot(snap, year);
+      });
+
+      setAdvertisers(newAdvertisers);
+    });
+  }, []);
+
+  return advertisers;
+}
```
kirakira0, line 107: Have you considered async functions? :)

client/src/components/Visualization/charts/ScatterPlot.js
kirakira0, line 28: Consider putting this function in a utility class since it isn't directly related to the Scatter Plot and can be generalized for other visualization components

scarletnguyen13: in addition, if your method is different from mine in Utils.js, just move this function there

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+ * json object that can be charted without further alteration.
+ * @param {object} snap a firestore snap object
+ * @param {string} year year corresponding to the relevant year. 
+ * @return {object}
+ */
+function formatAdvertiserCountSnapshot(snap, year) {
+  const numberOfAds = snap.data().numberOfAds;
+  const advertiser = snap.id;
+
+  const victoryFormattedAdvertiserCount =
```
scarletnguyen13, line 57: nit:

> const victoryFormattedAdvertiserCount = {
>     x: parseInt(year), 
>     y: parseInt(numberOfAds), 
>     fill: stringToColor(advertiser),
>     label: `${advertiser} had ${numberOfAds} ads in ${year}.`
> };

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+            .limit(queryLimit)
+            .get()
+            .then((snapshots) => {
+      const newAdvertisers = snapshots.docs.map(snap => {
+        return formatAdvertiserCountSnapshot(snap, year);
+      });
+
+      setAdvertisers(newAdvertisers);
+    });
+  }, []);
```
scarletnguyen13, line 104: the array should be `[advertisers]` instead of empty, because you would update the value of `advertisers` within the `useEffect`

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+
+  return advertisers;
+}
+
+const ScatterPlot = () => {
+  const queryLimit = 4; 
+
+  const advertisers2018 = useAdvertisers("2018", queryLimit);
```
scarletnguyen13, line 112: this seems like hard-coded, can you make an array of ["2018", "2019", "2020"] and map over it to create the `adverstisers` arrray on line 116 below?

client/src/components/Visualization/charts/ScatterPlot.js
rob-marcus: 80 char limit.

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+  
+  return (
+    <VictoryChart
+      theme={ VictoryTheme.material }
+      domain={{ x: [2017, 2021], y: [range.min*.5, range.max*2] }}
+      scale={{ y: "log" }}
+    >
+      <VictoryLabel
+        text={ chartTitle }
+        x={ 175 } y = { 30 }
```
scarletnguyen13, line 131: magic numbers should be stored in constants

client/src/components/Visualization/charts/ScatterPlot.js
rob-marcus: Our database only uses ad words so...

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -0,0 +1,156 @@
...
+        textAnchor="middle"
+      />
+
+      <VictoryAxis
```
scarletnguyen13, line 135: why can't this be on one line?

client/src/components/Visualization/charts/ScatterPlot.js
kirakira0: Ah ok, the line before had over 80. Different for links?

rob-marcus: Yup. You don't want to break up links bc then when you try and click on them it won't work. Same for searching for code etc. Adds unnecessary complexity.

client/src/components/Visualization/charts/ScatterPlot.js
rob-marcus: See [react rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). 

Basically, I had to choose between using a DRY `useAdvertiser/useEffect` and a somewhat not DRY `advertisers` array, vs a much less dry `useEffect` and a dry `advertisers` array. Since the latter is more lines and harder to implement, I chose the former.

client/src/components/Visualization/charts/ScatterPlot.js
rob-marcus: Did not realize that method existed! Neat, and have moved my method to be with it. 

Pros/cons of using `stringToColor`: 
pro: consistency without effort. Every time the user revisits/refreshes the chart with random color generator, all the bubbles will assume a different color. This could be confusing/misleading, i.e., did the data just change?, or, if it didn't, why did the chart change...? 

con: computationally more complex. The function is `O(n)` for string length `n`, random color is `O(1)`. Though practically none of the strings that I'd be calling it with are very large, don't think I've seen an advertiser with more than 30-40 characters in it. So it does incur a complexity cost but then again, not really.  

`m` strings, `O(mn)` cost to get all colors with my method. `O(m)` with scarlets method. But also I wouldn't want to use scarlets method without some kind of dict mapping advertisers to a color (would not want `adv_a 2019` to have a different color than `adv_a 2020`.) 

So additional client side space cost of `O(m)`, and complexity of adding that logic to the code. 

| Cost | Deterministic | Random | 
|------|--------------|----------|
|Comp| `O(mn)`        | `O(m)`    |
|(additional) space | `O(1)` | `O(m)`|

client/src/components/Visualization/charts/ScatterPlot.js
rob-marcus: Async in a hook requires you pull out the query into another function, introducing another layer of abstraction etc. Since this isn't even chaining, it's just thening, I think it's ok : )

PR #26, opened 2020-07-22
Connecting frontend to backend

scarletnguyen13:


client/src/components/SentimentAnalysis/Helpers.js
```diff
@@ -8,12 +8,14 @@ const AnalysisInput = props => {
```
rob-marcus, line 8: Please add a JSDoc somewhere to this with background on what this is doing/why/general context.

client/src/components/SentimentAnalysis/SentimentAnalysis.js
```diff
@@ -1,32 +1,53 @@
...
 
-  const {
-    mockContentTerms,
-    mockHeaderTerms,
-    mockHeaderScore,
-    mockHeaderMagnitude,
-    mockContentScore,
-    mockContentMagnitude,
-  } = MOCK_DATA;
+  const submitForm = e => {
```
rob-marcus, line 27: `e` is a bad name.

client/src/components/SentimentAnalysis/Helpers.js
kirakira0: +1

scarletnguyen13: It's just a textarea so I thought it's already self-explainatory :-?

rob-marcus: I mean in reference to the entire doc. The title `helpers` doesn't really tell us what's in the code, and it's good style to just have a JSdoc giving an overview without having to dissect the code/what is being returned.

PR #29, opened 2020-07-23
Fixed memory leak

rob-marcus:
**UPDATE**
- I have isolated the tab crashing error to geocharts. The following is still an error, but was not the root cause for the tabulating error. 

**Production problem:** 
- Going to `Ads Search` or `Sentiment Analysis` from `Data Visualization` tab will yield the following error: 
```
react-dom.production.min.js:196 TypeError: n is not a function
    at react-dom.production.min.js:199
    at e.unstable_runWithPriority (scheduler.production.min.js:18)
    at Wi (react-dom.production.min.js:120)
    at yu (react-dom.production.min.js:199)
    at bu (react-dom.production.min.js:204)
    at bs (react-dom.production.min.js:249)
    at e.unstable_runWithPriority (scheduler.production.min.js:18)
    at Wi (react-dom.production.min.js:120)
    at gs (react-dom.production.min.js:244)
    at os (react-dom.production.min.js:223)
```
**Bug solution:**
- I was unable to replicate this particular error or crash in a test environment when rendering only the scatter plot. But replicating the journey yielded a consistent memory leak error in the dev env.: 
```
index.js:1 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in ScatterPlot (at Visualization.js:18)
    in div (at Visualization.js:30)
    in div (created by Context.Consumer)
    in ScreenClassResolver (created by Col)
    in Col (at Visualization.js:29)
    in div (created by Row)
    in Row (at Visualization.js:27)
    in div (created by Context.Consumer)
    in ScreenClassProvider (created by Context.Consumer)
    in ScreenClassResolver (created by Container)
    in Container (at Visualization.js:25)
    in div (at Visualization.js:23)
    in Visualization (at header_constants.js:10)
```
- I'm guessing that this is the issue, and the dev env. just isn't crashing as fast due to probably a lot more memory available. 

Anyway, this was a pretty simple fix for the dev error. If I'm right that the two errors are connected than this should solve the tabbing problem. Otherwise still good to get rid of a memory leak. 

**Why does this fix work?** 
- I believe having `advertisers` in the array forces the `useEffect` function to essentially listen/check for updates on `advertisers`. If an update is detected, `useEffect` executes again. Since `advertisers` is being updated by `useEffect`, this creates a cycle and chews through memory (maybe. don't really understand memory leaks or the prod error, couldn't find a lot of information about it as it pertains to react hooks, but it seems reasonable.) 
- Removing `advertisers` removes the cycle, so no more memory leak. 

**How to avoid a repeat in the future?**
- Test prod. 

Note: there's a separate error that follows the pattern `Error: <line> attribute y1: Expected length, "NaN".` I'm still working on that, and I don't believe it's related to the page crashing.


client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -86,7 +86,7 @@ function useAdvertisers(year, queryLimit) {
 
         setAdvertisers(newAdvertisers);
       });
-  }, [advertisers, year, queryLimit]);
```
ailemeo, line 89: So what exactly is the variable used for?

rob-marcus: tl;dr: `advertisers` is a state variable tracking the current state of `advertisers` data loaded in from the API, and is set by `setAdvertisers` in `useEffect` . This variable is then returned by `useAdvertisers`.  

[react state hook](https://reactjs.org/docs/hooks-state.html)

ailemeo: I get it now. useState gives [currentState, functionToOverwriteTheCurrentState]

PR #30, opened 2020-07-23
Fix NaN victory error

rob-marcus:
**Production problem:**
- Loading `Data Visualization` will cause a non-fatal error with the following pattern: `Error: <line> attribute y2: Expected length, "NaN".`
- The error is caused because react initially loads the chart without any data (i.e., it loads the chart while asynchronously waiting for `firestore` data to load from `useEffect`, and once that data has been loaded, re-renders the chart contents.) 
- `getChartRange` is then given an empty list with no fields, so returns an undefined range. 
- These undefined values are then used to set the chart range, causing the victory error. 

**Bug solution:**
- Add a check to `getChartRange` to return the trivial range `{ min: 0, max: 1 }` if the input list is empty. 
   - `{ min: 0, max: 0 }` causes an error in victory because range can't be empty basically. 

I also rewrote the `useEffect` method with async/await for clarity as I thought this might initially be part of the bug. Alas, it isn't, but now it's `cleaner`.


client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -49,14 +49,19 @@ function formatAdvertiserCountSnapshot(snap, year) {
...
  * `https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax` 
+ * If the list is empty a range [0,1] will be returned. 
  * @param {List[object]} victoryJsonList a list of objects in a victory-ready 
  *     format.
  * @returns {object}
  */
 function getChartRange(victoryJsonList) {
-  const ranges = victoryJsonList.map(json => json.y);
+  if (victoryJsonList.length === 0) {
+    return {min: 0, max: 1};
```
scarletnguyen13, line 59: nit: `{ min: 0, max: 1 };`

client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -49,14 +49,19 @@ function formatAdvertiserCountSnapshot(snap, year) {
  * Returns the exclusive bounds for the range of all data points being charted.
  * Uses the spread `...` operator.
  * `https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax` 
+ * If the list is empty a range [0,1] will be returned.
```
kirakira0, line 54: Is the empty list an excepted behaviour? Why are you returning a range instead of throwing an error?

rob-marcus: Expected behavior. See PR doc for background.

PR #33, opened 2020-07-29
Word cloud

rob-marcus:
image

Background: 
- One bit of feedback from UX demo suggested a word cloud feature could be interesting. 
- Two basic ways to implement word cloud. 
  - Part of data visualization?
  - Part of search page? 
  - Each has +'s and -'s.
- Ultimately chose to integrate with search page. Why? 
   - Let's look at what word cloud feature would look like under data visualization: 
     - Short time frame to develop, has to be somewhat simple. But... 
     - Much more complex/less fluid user integration/interactivity.
       - User customizability can only be as complex as a guaranteed successful firestore query (i.e., not very complex at all, somewhat hard coded)
          - Maybe something like `see word cloud for ads from $state_x`. Anything else would get too complicated, fast. 
          - So size of user customization space is quantifiably small     
      - How to handle word cloud population? Could either do ahead of time (cloud funcs) or just in time (100% client side, big queries, slow)
         - Either way, a lot of infrastructure stuff to develop to enable something fairly limited in scope/user utility. 
   - Doing word cloud with the search page: 
     - Massive user customization and interactivity.
     - Faster than querying Firestore with data visualization (no additional querying being done, just reading from `DOM` eod.)   
     - Generate word cloud from first page of search results.
        - Make a word cloud of the most useful data (how many people care about page 2 of google? page 3? ...)
     - Still, why only first page? Why not all query responses? 
        - Basically `InstantSearch` frameworks is extremely aggressive with abstracting search results away.
           - Slightly OT, but further impact of this abstraction is that the cleanest and best way to retrieve search results is through the Dom, **not** through the search results code. 
        - Only dealing with what the search returns and is immediately available is the fastest, cleanest, most focused way of doing this. 
        - Maybe in future this would be a reason to consider *not* using `InstantSearch` framework. Overall small tradeoff IMO though.   
      - Another counterpoint: word cloud is good for ingesting lots of data though, no real need to limit it to a small subset. Might miss something? 
           - From personal experimentation, it doesn't really impact the end product. 
           - Played around in a python notebook, and more data doesn't mean better or more comprehensive word cloud. 
           - Basically big advertisers will completely drown out small advertisers, so you end up with something very focused anyway, but also with what appears to be lots of random noise in the background : (. 
           - Maybe could normalize scores or things with something more advanced than just a counter, e.g., tf-idf, but that's very out of scope for right now.  
- EOD, integrating with search is faster and better.


client/src/components/WordCloudModal/GenerateWordCloudValues.js
```diff
@@ -0,0 +1,88 @@
...
+                              'in',
+                              'is',
+                              'it',
+                              'its',
+                              'of',
+                              'on',
+                              'the',
+                              'this',
+                              'to',
+                              '|']);
```
kirakira0, line 28: How did you decide on these words in particular? You have "this" but not "that",  "its" but not "it's", and some conjunctions but not others.

client/src/components/WordCloudModal/GenerateWordCloudValues.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Generate the values of a react word cloud based on the current 
+ * algolia search results in the DOM.
+ * Word cloud format is [{ text: word0, value: number0 }, ...] 
+ * @return {Object}
+ */
+const generateWordCloudValues = () => {
+  const searchResultsInnerTextList = 
+      Array.from(document.getElementsByClassName('ais-Hits-item'))
```
kirakira0, line 74: Consider leaving a comment explaining 'ais-Hits-item'

client/src/components/WordCloudModal/WordCloudModal.js
```diff
@@ -0,0 +1,58 @@
...
+import ReactWordcloud from 'react-wordcloud';
+
+import generateWordCloudValues from './GenerateWordCloudValues';
+
+Modal.setAppElement('#root');
+
+const wordCloudOptions = {
+    colors: ['#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd', '#8c564b'],
+    enableTooltip: true,
+    deterministic: false,
```
kirakira0, line 12: Looks like you were going for an alphabetized list, so switch the order of these :)

client/src/components/Search/Search.css
```diff
@@ -21,7 +21,19 @@
   left: 20px;
 }
 
+.button-left {
```
scarletnguyen13, line 24: nit: for better naming, rename to `position-left` and `position-right`

client/src/components/WordCloudModal/GenerateWordCloudValues.js
```diff
@@ -0,0 +1,88 @@
...
+ *              getting search results from the DOM
+ *              and counting instances of individual or hyphenated words
+ *              using a dictionary mapping approach for efficiency.   
+ *              Un-interesting words are filtered from the word cloud.
+ * Date: 7/28
+ * Author: Rob Marcus
+ */
+
+// Set of invalid word cloud words/strings. 
+const INVALID_WORDS = new Set(['',
```
scarletnguyen13, line 13: Maybe store this in a constant file?

client/src/components/WordCloudModal/GenerateWordCloudValues.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Make a map relating string words (keys)  
+ * to word cloud formatted json objects (values)
+ * Word cloud format is [{ text: word0, value: number0 }, ...]
+ * For speed, wordMap format is {word0: { text: word0, value: number0 }, ... }
+ * @param {list[string]} wordList search results split space wise
+ * @return {Object}
+ */
+function makeWordMap(wordList) {
```
scarletnguyen13, line 47: nit: `createWordMap` is a more common naming convention - we usually use `create` instead of `make`

client/src/components/WordCloudModal/GenerateWordCloudValues.js
```diff
@@ -0,0 +1,88 @@
...
+      searchResultsInnerTextList.filter(word => isWordValid(word));
+
+  const searchResultsWordMap = makeWordMap(searchResultsFilteredWordList);
+
+  const wordCloudValues = Object.values(searchResultsWordMap);
+  
+  return wordCloudValues;
+}
+
+export default generateWordCloudValues;
```
scarletnguyen13, line 88: if export a function instead of a React component, use `export { generateWordCloudValues };`

client/src/components/WordCloudModal/GenerateWordCloudValues.js
rob-marcus: Experimentally looked at the word cloud and took notes on what words were polluting the chart basically. Have added a clarifying comment.

client/src/components/WordCloudModal/GenerateWordCloudValues.js
```diff
@@ -0,0 +1,74 @@
...
+  wordList.forEach(word => {
+
+    // Remove all non alphanumeric or $-sign characters (i.e., punctuation)
+    // for comprehensive string matching without losing context.
+    const filteredWord = word.replace(/\W$/g, ''); 
+    
+    if(wordMap.hasOwnProperty(filteredWord)) {
+      wordMap[filteredWord].value += 1;
+    } else {
+      wordMap[filteredWord] = { text: filteredWord, value: 1 };
```
andrewlathamgoogle, line 45: text property is redundant since it's also the key

client/src/components/WordCloudModal/GenerateWordCloudValues.js
```diff
@@ -0,0 +1,74 @@
...
+ * @return {Object}
+ */
+function createWordMap(wordList) {
+  const wordMap = {};
+
+  wordList.forEach(word => {
+
+    // Remove all non alphanumeric or $-sign characters (i.e., punctuation)
+    // for comprehensive string matching without losing context.
+    const filteredWord = word.replace(/\W$/g, '');
```
andrewlathamgoogle, line 40: Check if it's blank.
Also you can put isWordValid after this to get rid of words that become invalid words after this substitution.

client/src/components/WordCloudModal/GenerateWordCloudValues.js
rob-marcus: This is a really convenient way of packaging the data for the word cloud format so no further processing is done on the map, just return the map values as a list and it's packaged. I originally did do it how you're suggesting plus some further processing to format again, but this is a fair bit cleaner I think and only a tiny bit non-repetitive.

PR #38, opened 2020-07-30
Similar Ads Feature

kirakira0:
image


client/src/components/SentimentAnalysis/SimilarAds.js
```diff
@@ -0,0 +1,59 @@
...
+  const contentSentiment = `{"score": ${props.contentScore}, "magnitude": ${props.contentMagnitude}}`;
+
+  const [results, setResults] = useState([]);
+  let resultItems = [];
+
+  useEffect(() => {
+    async function fetchSimilarAds() {
+      
+      let similarAdsPromises = [];
+      similarAdsPromises.push(database.collection(COLLECTION)
```
rob-marcus, line 17: Why push this when you can just define the array with it by default, e.g., 
```javascript
let similarAdsPromises = 
    [database.collection(COLLECTION)
             .blah
    ];
```

client/src/components/SentimentAnalysis/SimilarAds.js
```diff
@@ -0,0 +1,59 @@
...
+            .where('headlineSentiment', '==', headlineSentiment)
+            .where('contentSentiment', '==', contentSentiment)
+            .limit(10)
+            .get()
+      ); 
+      let similarAds = await Promise.all(similarAdsPromises);
```
rob-marcus, line 23: Since there is only one promise being awaited you don't really need `Promise.all`.

client/src/components/SentimentAnalysis/SentimentAnalysis.js
```diff
@@ -149,6 +149,11 @@ const SentimentAnalysis = () => {
     setIsModalOpen(true);
   };
 
+  function toString(sentimentValue) {
```
rob-marcus, line 152: since `toString` is already a built in class method, perhaps a different name for clarity?

client/src/components/SentimentAnalysis/SentimentAnalysis.css
```diff
@@ -80,3 +80,52 @@
 .bottom-padding {
   padding-bottom: 40px;
 }
+
+.similar-container {
```
scarletnguyen13, line 84: nit: alphabetizing

client/src/components/SentimentAnalysis/SentimentAnalysis.css
```diff
@@ -80,3 +80,52 @@
...
+  transition: 0.5s;
+}
+
+.similar-ad a:hover {
+  color: black;
+  transition: 0.5s;
+}
+
+h4, #card-description {
+  text-align: center;
```
scarletnguyen13, line 125: Andrew will fight hard for this, just saying lol

client/src/components/SentimentAnalysis/SentimentAnalysis.js
scarletnguyen13: how about `convertSentimentToString(value)`?

client/src/components/SentimentAnalysis/SimilarAds.js
```diff
@@ -0,0 +1,59 @@
...
+import React, { useState, useEffect } from 'react';
+
+const SimilarAds = props => {
+
+  const COLLECTION = 'ads';  
+  const headlineSentiment = `{"score": ${props.headlineScore}, "magnitude": ${props.headlineMagnitude}}`;
+  const contentSentiment = `{"score": ${props.contentScore}, "magnitude": ${props.contentMagnitude}}`;
+
+  const [results, setResults] = useState([]);
+  let resultItems = [];
```
scarletnguyen13, line 11: We can put this locally inside the `useEffect` right before its usage - also, use `const` for `resultItems` since it doesn't reference to another object later

server/src/main/java/com/google/sps/utils/ReportReader.java
```diff
@@ -24,9 +24,9 @@
   private static final String CSV_FILE_PATH = "/output.csv";
   // NUMBER_CSV_HEADER_ROWS helps the parser know what rows to skip when reading the CSV.
   private static final int NUMBER_CSV_HEADER_ROWS = 2;
-  private static final String COLLECTION = "ads";
-  private static final int START_ROW_INDEX = 20001;  // First row sent to processor (inclusive).
-  private static final int END_ROW_INDEX = 25000; // Last row sent to processor (inclusive).
+  private static final String COLLECTION = "dev_ads";
```
scarletnguyen13, line 27: May I ask why you changed this to `dev_ads` instead?

client/src/components/SentimentAnalysis/SentimentAnalysis.css
kirakira0: Haha that's ok I'm not super attached to it. If there's a better alternative I'll gladly use that instead

client/src/components/SentimentAnalysis/SentimentAnalysis.js
```diff
@@ -149,6 +149,11 @@ const SentimentAnalysis = () => {
     setIsModalOpen(true);
   };
 
+  function convertSentimentToString(sentimentValue) {
+    let str = sentimentValue.toString();
+    return (str.includes('.') ? str : str + '.0');
```
andrewlathamgoogle, line 154: What's going on here?  Please add a comment explaining.

client/src/components/SentimentAnalysis/SimilarAds.js
```diff
@@ -0,0 +1,55 @@
...
+
+  const [results, setResults] = useState([]);
+
+  useEffect(() => {
+    async function fetchSimilarAds() {
+      let similarAds = await database
+          .collection(COLLECTION)
+          .where('headlineSentiment', '==', headlineSentiment)
+          .where('contentSentiment', '==', contentSentiment)
+          .limit(9)
```
andrewlathamgoogle, line 18: make this a constant

client/src/components/SentimentAnalysis/SimilarAds.js
```diff
@@ -0,0 +1,55 @@
...
+          .get();
+      const resultItems = [];
+      for (let ad of similarAds.docs) {
+        resultItems.push(
+          <div className="similar-ad">
+            <p><b>Headline</b>: {ad.data().headline}</p>
```
andrewlathamgoogle, line 24: This is the nitpickiest of nitpicks, but just as an educational fyi, when you use tags like \<b\> and \<strong\> SEO bots assign semantic meaning to them and think they're the most important parts of your website.  Since that isn't the case here, using a \<span\> with font-weight: 700 would technically be more correct.

This is true in general of tags -- HTML tags represent semantic meaning.  There is a pretty big difference between \<h1\> and \< span style="font-size: 32px, font-weight: 700" \> in terms of how bots and rendering/interpreter tools will interpret your website.

client/src/components/SentimentAnalysis/SimilarAds.js
```diff
@@ -0,0 +1,55 @@
...
+      setResults(resultItems);
+    } 
+    fetchSimilarAds();
+  }, [])
+
+  return (
+    <div className="similar-container">
+      <div className="similar-card">
+        <h4>Similar Advertisements By Sentiment</h4>
+        <p id="card-description">
```
andrewlathamgoogle, line 42: would this look better as a list?

client/src/components/SentimentAnalysis/SentimentAnalysis.css
andrewlathamgoogle: The alternative is \<div style="display: flex; justify-content: center;"\>\<span>My text\>\</div\>.

For a header like this, text-align is fine.

PR #40, opened 2020-07-30
Added more tick values and y label padding

rob-marcus:
Title. New look: 

image
Added ticks for the min/max advertiser ad count for every year, so 6 ticks total now. 

Also, since Warren went so heavy with ad words ads, I added a zoom component so it's easier to explore everyone else's performance.

### DIFF
--- a/client/src/components/Visualization/charts/ScatterPlot.js
+++ b/client/src/components/Visualization/charts/ScatterPlot.js
@@ -114,7 +114,6 @@ const ScatterPlot = () => {
   // Get the min and max tick values for each year in the chart.
   advertisers.forEach(year => {
     const range = getChartRange(year);
-
     ranges.push(range.min);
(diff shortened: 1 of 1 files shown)


client/src/components/Visualization/charts/ScatterPlot.js
```diff
@@ -102,23 +105,34 @@ function useAdvertisers(year, queryLimit) {
...
-                       ...advertisers2019,
-                       ...advertisers2020];
+  const ranges = [];
 
-  const range = getChartRange(advertisers);
-  const chartTitle = `T${queryLimit} Most prolific ad words advertisers/year`;
+  // Get the min and max tick values for each year in the chart.
+  advertisers.forEach(year => {
+    const range = getChartRange(year);
+
```
andrewlathamgoogle, line 117: -newline

